PR #11, opened 2019-10-11
fixing infinite loop in the script

algonautshant:
when an argument is passed which does not match any of the cases, the script will go to an infinite loop.
Adding a default case to consume the argument.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -86,7 +86,8 @@ else
                 shift
                 ;;
             *)
-                shift
+                echo "Unsupported flag: " $1
+                exit 1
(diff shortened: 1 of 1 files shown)


scripts/test.sh
```diff
@@ -85,6 +85,9 @@ else
                 cross=true
                 shift
                 ;;
+            *)
+                shift
+                ;;
```
tsachiherman: This would be triggered when one passes an argument which isn't one of the supported ones, right ?
In this case, I think that we want to print and error message followed by "exit 1".

PR #13, opened 2019-10-14
change dependency to go-algorand master

mjiang102628:


scripts/setup.sh
```diff
@@ -65,14 +65,17 @@ else
...
-# this is the link for linux; change this if on mac or windows
-curl -L https://github.com/algorand/go-algorand-doc/blob/master/downloads/installers/linux_amd64/install_master_linux-amd64.tar.gz?raw=true -o ~/inst/installer.tar.gz
-tar -xf ~/inst/installer.tar.gz -C ~/inst
-~/inst/update.sh -i -c stable -p ~/node -d ~/node/data -n
+# # test last release
+# mkdir ~/inst
+# # this is the link for linux; change this if on mac or windows
+# curl -L https://github.com/algorand/go-algorand-doc/blob/master/downloads/installers/linux_amd64/install_master_linux-amd64.tar.gz?raw=true -o ~/inst/installer.tar.gz
+# tar -xf ~/inst/installer.tar.gz -C ~/inst
+# ~/inst/update.sh -i -c stable -p ~/node -d ~/node/data -n
```
EvanJRichard: Total nit: Could maybe make a switch between test latest/ test last release? or change comment to like "uncomment this, and comment above block, to test previous release rather than latest"?

network_config/template.json
```diff
@@ -1,5 +1,6 @@
 {
     "Genesis": {
+        "ConsensusProtocol": "future",
```
EvanJRichard: We want `future` always, right? I think so, but wanted to make sure

PR #32, opened 2019-12-06
Add new decimal param to java methods.

winder:


java_cucumber/src/test/java/java_cucumber/Stepdefs.java
```diff
@@ -963,6 +963,7 @@ public void default_asset_creation_transaction_with_total_issuance(Integer asset
                 acl.transactionParams().getGenesisID(),
                 new Digest(acl.transactionParams().getGenesishashb64()),
                 BigInteger.valueOf(assetTotal),
+                BigInteger.valueOf(0),
```
algonautshant: Perhaps we use int instead of BigInteger?

PR #39, opened 2020-01-30
Run tests in a docker container.

winder:
Run cucumber tests in a docker container.
Make setup.sh more configurable (`scripts/docker/setup.py`), needs some more work to allow a way to pull down production binaries.

For now you can use `run_docker.sh` to build/execute the tests. You can also pass flags to test.sh to configure the tests as usual, for example once the image is built run your local code in the container mount your source directory and tell the setup script where to find the mounted directory:
```
docker run -it \
      -v $(pwd):/opt/sdk-testing \
      -v /home/will/algorand/java-algorand-sdk:/opt/mounted_java_sdk \
      sdk-testing:latest \
      /bin/bash -c "\
            GO111MODULE=off && \
            ./scripts/docker/setup.py --java-dir /opt/mounted_java_sdk && \
            ./scripts/docker/test.sh --java"
```


scripts/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM ubuntu:18.04
+
+ENV DEBIAN_FRONTEND noninteractive
+
+# Basic dependencies
+ENV HOME /opt
```
bricerisingalgorand: I would leave ENV var settings for the end of the Dockerfile, so that when you decide to change them you don't have to redownload all of your dependencies each time. A good rule of thump is that you want order steps by their likelihood to change, so you can better take advantage of the docker cache

run_docker.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+rootdir=`dirname $0`
+pushd $rootdir
+docker build -t sdk-testing -f ./scripts/docker/Dockerfile $rootdir
```
bricerisingalgorand: It would be better for us to put this in a docker registry than have people build it every time. This will make tests take longer to run and create a blocker when the dockerfile breaks (which it will eventually since people update packages, change urls and stuff)

scripts/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+RUN apt-get update && apt-get install -y apt-utils curl git git-core bsdmainutils
+
+# Install python dependencies
+ENV PYENV_ROOT $HOME/pyenv
```
bricerisingalgorand: This should probably be replaced with something like this

```
RUN apt-get update && \
    apt-get install -y \
    build-essential \
    libssl-dev \
    libffi-dev \
    python3-dev \
    python3-pip \
```

scripts/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    && pip3 install behave -q
+
+# Install Javascript dependencies
+RUN curl -sL https://deb.nodesource.com/setup_13.x | bash -
+RUN apt-get install -y nodejs
+
+# Install Java dependencies
+RUN apt-get install -y maven default-jdk
+
+# Install Go dependencies
```
bricerisingalgorand: Why not use the golang official image as a base instead of installing all this on ubuntu?

scripts/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+ARG GOLANG_VERSION=1.12.16
+RUN curl https://dl.google.com/go/go${GOLANG_VERSION}.linux-amd64.tar.gz -o $HOME/go.tar.gz
+RUN tar -xvf $HOME/go.tar.gz -C /usr/local
+ENV GOROOT /usr/local/go
+ENV GOPATH $HOME/go
+ENV PATH $GOROOT/bin:$PATH
+
+RUN mkdir -p $HOME/sdk-testing
+WORKDIR $HOME/sdk-testing
+CMD ["/bin/bash"]
```
bricerisingalgorand: why not make this `CMD "GO111MODULE=off && ./scripts/setup.sh && ./scripts/test.sh"`? That way those that just want to run the test don't need to know the contents of the docker image to run them and just do:

```
docker run -v $(pwd):/opt/sdk-testing sdk-testing:latest
```

run_docker.sh
```diff
@@ -0,0 +1,6 @@
...
+docker run -v $(pwd):/opt/sdk-testing sdk-testing:latest /bin/bash -c "GO111MODULE=off && ./scripts/setup.sh && ./scripts/test.sh"
```
bricerisingalgorand: I would add a `-ti` option to this so that a user can exit the test more easily if they want to

scripts/docker/Dockerfile
winder: I use $HOME throughout the script so I put this one at the top

scripts/docker/Dockerfile
winder: atp-get updat runs a couple lines earlier

scripts/docker/Dockerfile
winder: I thought about that, but you could say the same about java/javascript/python. In terms of complexity python is the most involved installation at the moment.

scripts/docker/Dockerfile
winder: I like that idea for the default case and will make the change, but I'm sort of imagining this as a runtime environment. In most cases what is now the `test.sh` command will have a number of additional options and/or more volumes mounted.

run_docker.sh
winder: :+1:

scripts/docker/Dockerfile
bricerisingalgorand: I meant that it could be better to install python3 and then do pip3 installs for any dependencies we need, unless there's a reason why we need to use pyenv

winder:
I'm going to merge this in so that I can use it for template testing, it shouldn't have any effect to existing behaviors, we can do that later if we want to make this the preferred way to run the cucumber tests.

PR #44, opened 2020-02-10
fix go sdk installation

mjiang102628:
- add missing exclamation mark
- switched default branch of go-algorand-sdk to develop (not a change in this repo)
- change godog path to github.com/cucumber/godog due to moved package

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -25,7 +25,6 @@ go get github.com/DATA-DOG/godog/cmd/godog
 if ! $go
 then
     go get -u github.com/algorand/go-algorand-sdk/...@develop
-    go generate github.com/algorand/go-algorand-sdk/...
 fi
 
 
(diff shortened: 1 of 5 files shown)


scripts/setup.sh
```diff
@@ -21,12 +21,13 @@ case "$1" in
         ;;
 esac
 
-go get github.com/DATA-DOG/godog/cmd/godog
-if $go
+
+if ! $go
```
EvanJRichard: Am I reading this right, that the script calls `go get go-algorand-sdk` if `--go` is _not_ supplied, and skips `go get go-algorand-sdk` if it is supplied?

mjiang102628: Yes; because setup is called from sdkupdate, and specifying some flag tells the script that that's the one that we already have a local version of.

PR #52, opened 2020-03-17
Fix bug where master was being targeted for go get rather than develop

EvanJRichard:
This simple hotfix changes  `develop` to be the target of `setup.sh: go get go-algorand-sdk`. This is needed for tests to pass.


scripts/setup.sh
```diff
@@ -83,6 +83,6 @@ source "$SCRIPTPATH/config_future"
 
 mkdir ~/inst
 # this is the link for linux; change this if on mac or windows
-curl -L https://algorand-releases.s3.amazonaws.com/channel/stable/install_stable_linux-amd64_2.0.4.tar.gz -o ~/inst/installer.tar.gz
+curl -L https://algorand-releases.s3.amazonaws.com/channel/stable/install_stable_darwin-amd64_2.0.4.tar.gz -o ~/inst/installer.tar.gz
```
rotemh: Consider letting the script check what OS is running on and choosing the link automatically. No need to change it manually everytime

EvanJRichard: @rotemh Agreed that we should do that eventually, it's in the backlog: https://github.com/orgs/algorand/projects/3#card-28255567 For now I will keep the PR to just the bugfix.

PR #57, opened 2020-04-14
Indexer Support

winder:
These changes require updating the SDK driver scripts.

To use these the workflow is as follows:
1. Start the test harness environment with `./scripts/up.sh`
2. Run the SDK tests (i.e. from the java SDK repo run `mvn test -Dskip.integration.tests=false -e`)

Thats it!

We should also maintain per-SDK docker containers to make it easy to include these tests in the CICD process. More on that in the Java SDK PR.


docker/algod/setup.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    # Reset network dir before creating a new one.
+    if os.path.exists(args.network_dir):
+        shutil.rmtree(args.network_dir)
+
+    # $BIN_DIR/goal network create -n testnetwork -r $NETWORK_DIR -t network_config/$TEMPLATE
+    subprocess.check_call(['%s/goal network create -n testnetwork -r %s -t %s' % (bin_dir, network_dir, template)], shell=True)
+    node_dir, kmd_dir = algod_directories(network_dir)
+
+    # TODO: Delete this?
```
EvanJRichard, line 75: should this be deleted?

docker/algod/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:19.10
+
+ENV DEBIAN_FRONTEND noninteractive
+
+# Basic dependencies
+ENV HOME /opt
+RUN apt-get update && apt-get install -y apt-utils curl git git-core bsdmainutils python3
+
+# Install algorand-sdk-testing script dependencies
+#RUN pip3 install gitpython
```
EvanJRichard, line 10: is this commented-out line still needed?

PR #65, opened 2020-04-30
Evan/minortweaks

EvanJRichard:
--- not a bug but I would like to change it if it’s okay with you guys: paths supplied in _paths.feature are not alphabetized, but AFAICT go can only do alphabetized queries. edited _paths.feature golden paths to use alphabetized queries
--- not a bug but I would like to change it if it’s okay with you guys: feature files use the string “none” to mean the empty string. changed these examples to the literal empty string.
--- fixed a bug: the golden value for rewards pool in the GetBlock matched the golden _json_ not the golden _b64_. edited expected rewards pool to reflect b64 (thanks Michelle for noticing this)
--- fixed a bug: in one feature file where an argument “AfterAddress” was defined but there was no slot for it in examples
--- fixed a bug/QoL improvement: the empty jsons were literal empty files, but not valid jsons. changed them to be valid empty jsons.
--- tagged indexer_paths and indexer_responses with `@indexer`


winder:
> --- not a bug but I would like to change it if it’s okay with you guys: paths supplied in _paths.feature > are not alphabetized, but AFAICT go can only do alphabetized queries. edited _paths.feature >golden paths to use alphabetized queries

We might not be able to do these tests if the order of fields changes, I think it's probably unreasonable to require the SDKs to create the path in a specific order. Presumably the python SDK isn't alphabetized.

>--- not a bug but I would like to change it if it’s okay with you guys: feature files use the string “none” >to mean the empty string. changed these examples to the literal empty string.

+1. Assuming the tests are written without special distinction between "none" and "".

> --- fixed a bug: in one feature file where an argument “AfterAddress” was defined but there was no >slot for it in examples

`AfterAddress` was renamed to `next`, and all endpoints that support paging call it `next` now. Previously there were slightly different names which reflect implementation details (i.e. the `next` token is an address, thus `AfterAddress`)

>--- tagged indexer_paths and indexer_responses with @indexer

:+1: 

> --- fixed a bug/QoL improvement: the empty jsons were literal empty files, but not valid jsons. changed them to be valid empty jsons.

Are these used for error cases? In theory the server could return a status code + empty response.

>--- fixed a bug: the golden value for rewards pool in the GetBlock matched the golden json not the >golden b64. edited expected rewards pool to reflect b64 (thanks Michelle for noticing this)

EvanJRichard:
_@winder wrote:_
>We might not be able to do these tests if the order of fields changes, I think it's probably unreasonable to require the SDKs to create the path in a specific order. Presumably the python SDK isn't alphabetized.
Gotcha, I'll change the go-sdk test impl to split up the passed path and compare it regardless of order

>Are these used for error cases?
Not at this time, just for the cases where a response isn't expected.

winder:
Talked with @EvanJRichard this morning about the compound path tests. The order of path parameters is an implementation detail so we wanted to find a way to avoid having it be an implementation requirement.

The solution: only set one parameter at a time. We can have all the tests we need to validate that path/parameter names are correct and values are encoded properly without worrying about any implementation details.

Concrete example:
before
```
    Examples:
      |path | max| format |
      |/v2/transactions/pending?format=msgpack       | 0  | msgpack |
      |/v2/transactions/pending?max=1&format=msgpack | 1  | msgpack |
      |/v2/transactions/pending?format=msgpack&max=1 | 1  | msgpack |
```
after
```
    Examples:
      |path | max| format |
      |/v2/transactions/pending?format=msgpack | 0  | msgpack |
      |/v2/transactions/pending?max=1          | 1  |         |
```

features/unit/v2algodclient_paths.feature
```diff
@@ -32,15 +32,15 @@ Feature: Algod REST API v2
     Examples:
       |path | max| format |
       |/v2/transactions/pending?format=msgpack | 0  | msgpack |
-      |/v2/transactions/pending?max=1&format=msgpack | 1  | msgpack |
+      |/v2/transactions/pending?format=msgpack&max=1 | 1  | msgpack |
```
winder, line 35: There are some more in this file

EvanJRichard: @winder I'd like to leave these particular ones as-is because `go` and `js` can only request msgpack, they cannot generate `?format=json` or `format`-omitted queries (this was requested back during design in March)

winder: Maybe we can have the convention be `format=msgpack` goes first. That can be the only artificial order restriction.

PR #69, opened 2020-05-06
remove registerpartkeys and shutdown tests

EvanJRichard:
Per review feedback in go-sdk and js-sdk, this changeset will remove registerpartkeys and shutdown tests.


EvanJRichard:
Merging without review as this was an action item generated by another review/approved in another PR.

PR #71, opened 2020-05-11
-- Rekey feature files

rotemh:


features/unit/rekey.feature
```diff
@@ -0,0 +1,28 @@
+@rekey
```
winder, line 1: Rename to `@unit.rekey`, then we can merge this in!

winder: This was causing some issues for SDK development so I went ahead updated the change.

PR #81, opened 2020-05-26
Rekey-aware indexer files

EvanJRichard:
This proposes to add rekey-to/auth-addr response and path checks. I have included them in the existing tags, but could split them out into new files for new tags.

I have written an update that makes these tests pass in go, javascript, and python.


features/unit/v2indexer_rekey_paths.feature
```diff
@@ -0,0 +1,33 @@
+@unit
```
winder, line 1: could you put this in `v2indexerclient_paths.feature`? You can tag each scenario with `unit.indexer.rekey`.

features/unit/v2indexer_rekey_responses.feature
```diff
@@ -0,0 +1,20 @@
+@unit
+@unit.indexer.rekey
```
winder, line 2: Same here, lets keep adding to `v2indexerclient_responses.feature`

features/unit/v2indexer_rekey_responses.feature
```diff
@@ -0,0 +1,20 @@
...
+Feature: Indexer Client v2 Responses
+  Scenario Outline: SearchForAccounts response, authorizing address
+    Given mock http responses in "<jsonfiles>" loaded from "<directory>"
+    When we make any SearchAccounts call
+    Then expect error string to contain "<err>"
```
winder, line 7: Remove the error check, we can add it back when we have errors.

features/unit/v2indexer_rekey_responses.feature
```diff
@@ -0,0 +1,20 @@
...
+    And the parsed SearchAccounts response should be valid on round <roundNum> and the array should be of len <len> and the element at index <index> should have authorizing address "<authaddr>"
```
winder, line 8: Add an improved `Then the re-serialized object equals the response in "<jsonfiles>" loaded from "<directory>"`

We can take the SDK object, serialize it back to JSON, and compare it to the input file.

This way we exactly compare all fields and don't need to re-implement the check for every test.

winder: In the Java SDK I was able to implement all response verify functions with the same helper:
```
  @Then("the parsed SearchForAssets response should be valid on round {long} "
      + "and the array should be of len {long} "
      + "and the element at index {long} "
      + "should have asset index {long}")
  public void the_parsed_SearchForAssets_response_should_be_valid_on_round_and_the_array_should_be_of_len_and_the_element_at_index_should_have_asset_index(
      Long int1, Long int2, Long int3, Long int4) throws IOException {
    verifyResponse(assetsResponse, shared.bodyFile);
  }
```

features/unit/v2indexer_rekey_paths.feature
EvanJRichard: Ah, I had syntax errors from improperly trying to tag just the scenarios, thought it was impossible - looks like I was just doing it wrong :+1:

features/unit/v2indexer_rekey_responses.feature
EvanJRichard: I'm not sure how easy it is to directly serialize/deserialize the jsons using the tools the SDKs expose to a user dev. I'll take a whack at it!

PR #82, opened 2020-05-26
Update the restart.sh script.

algonautshant:


scripts/restart.sh
```diff
@@ -2,14 +2,19 @@
 
 # This script is useful for development to quickly restart the
 # environment for another round of tests.
-set -e
 
-rootdir=`dirname $0`
-pushd $rootdir/.. > /dev/null
```
winder, line 8: The above lines are still needed, otherwise you can't run the script from outside the root directory.

winder: Shouldn't we keep `set -e` in case `docker-compose down` or `docker-compose rm --force` fail? Would there be any point trying to bring things back up?

algonautshant: Fixed. thanks.

PR #85, opened 2020-05-29
DRAFT: algorand-sdk-testing pipeline

bricerisingalgorand:
This is a draft PR for changes that will enable us to make a muleCI pipeline for this project


cicd/docker/algod/entrypoint.sh
```diff
@@ -0,0 +1,28 @@
...
+    python3 /algod/setup.py install \
+      --bin-dir "$HOME/node" \
+      --channel "${CHANNEL}"
+      --algod-version ${ALGOD_VERSION}
+  fi
+fi
+
+mkdir -p /opt/testnetwork
+
+python3 /algod/setup.py configure-network \
```
winder, line 18: This should be done at build time, why reconfigure the container every time it starts?

cicd/docker/algod/setup.py
```diff
@@ -15,32 +15,41 @@
...
 
-install = subparsers.add_parser('install', parents=[base_parser], help='Install the network.')
-install.add_argument('--network-template', required=True, help='Path to private network template file.')
-install.add_argument('--network-token', required=True, help='Valid token to use for algod/kmd.')
-install.add_argument('--algod-port', required=True, help='Port to use for algod.')
-install.add_argument('--kmd-port', required=True, help='Port to use for kmd.')
+install = subparsers.add_parser('install', parents=[base_parser], help='Install binaries.')
+install.add_argument('--bin-dir', required=True, help='Location to install algod binaries.')
+install.add_argument('--channel', required=True, help='Channel to install, nightly is a good option.')
+install.add_argument('--algod-version', required=False, default='2.0.4', help='Channel to install, nightly is a good option.')
```
winder, line 24: This doesn't need to be configurable, the "update" script will always pull down the most recent version. If that changes then we'll need to update more than just this script.

cicd/docker/algod/setup.py
```diff
@@ -15,32 +15,41 @@
...
-def setup_algod(bin_dir, channel):
+def setup_algod(bin_dir, channel, algod_version):
     """
     Download and install algod.
     """
     home = expanduser('~')
     os.makedirs("%s/inst" % home, exist_ok=True)
     print('downloading updater...')
-    url='https://algorand-releases.s3.amazonaws.com/channel/stable/install_stable_linux-amd64_2.0.4.tar.gz'
+    url=f"https://algorand-releases.s3.amazonaws.com/channel/stable/install_stable_linux-amd64_{algod_version}.tar.gz"
```
winder, line 50: If we wanted to simplify this, we could include the new and improved `update.sh` script in this repo. The new script automatically finds and downloads the latest updater.

cicd/docker/algod/setup.py
```diff
@@ -57,17 +66,7 @@ def algod_directories(network_dir):
...
-    """
-    Create a private network.
-    """
-    # Reset network dir before creating a new one.
-    if os.path.exists(args.network_dir):
-        shutil.rmtree(args.network_dir)
-
-    # $BIN_DIR/goal network create -n testnetwork -r $NETWORK_DIR -t network_config/$TEMPLATE
-    subprocess.check_call(['%s/goal network create -n testnetwork -r %s -t %s' % (bin_dir, network_dir, template)], shell=True)
+def configure_network(network_dir, token, algod_port, kmd_port):
```
winder, line 69: Why do we need to split the create and configure?

If you want it to be more general purpose, make the port/token optional fields and don't set them when they aren't provided.

cicd/docker/indexer/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+ENV DATABASE_PORT=5432
+ENV DATABASE_USERNAME=algorand
+ENV DATABASE_PASSWORD=harness
+
+ADD https://algorand-testdata.s3.amazonaws.com/indexer/algorand-indexer /tmp/algorand-indexer
+RUN chmod +x /tmp/algorand-indexer
+
+COPY cicd/docker/indexer/setup.sh /tmp/setup.sh
+COPY cicd/docker/indexer/e2edata.tar.bz2 /tmp/e2edata.tar.bz2
+COPY cicd/docker/indexer/init-scripts /docker-entrypoint-initdb.d
```
winder, line 14: the init-scripts are for the postgres container, not the indexer container.

docker-compose.yml
```diff
@@ -2,35 +2,26 @@ version: '3'
...
     networks:
       - sdk-harness
     environment:
       DATABASE_NAME: dataset1
-
-  indexer-db:
+  postgres:
     image: "postgres"
     container_name: sdk-harness-postgres
-    volumes:
```
winder, line 32: This is needed, it enables the `POSTGRES_MULTIPLE_DATABASES` environment variable which allows us to initialize multiple databases

docker-compose.yml
```diff
@@ -2,35 +2,26 @@ version: '3'
 
 services:
   algod:
-    image: "sdk-harness-algod"
+    image: algorand/algorand-sdk-testing-algod:latest
```
winder, line 5: If I need to update these dockerfiles, what is the process for testing the environment locally?

cicd/docker/algod/entrypoint.sh
```diff
@@ -0,0 +1,28 @@
+set -e
+
+if [ $UPDATE_ON_START = true ]; then
+  if [ -z ${ALGOD_VERSION} ]; then
+    python3 /algod/setup.py install \
```
winder, line 5: no need to run the full `setup.py` script here, update.sh/find-nodes.sh should do the trick:
```
$HOME/node/update.sh $($HOME/node/find-nodes.sh /opt/testnetwork)
```

Also the ALGOD_VERSION bit doesn't look useful so this can be collapsed to a single command

cicd/docker/algod/entrypoint.sh
bricerisingalgorand: This is so that if things like algod/kmd port or network token needs to be changed at start time. There can be instances where this is useful like times when we may want to run several of these on one kubernetes pod with shared networking

cicd/docker/algod/setup.py
bricerisingalgorand: Oh I thought when we talked one of the things you wanted was to be able to run any of the versions of algod that we've built. I could change it, but the way it's written now this can run a container with any version of algod we've published running a network for testing/troubleshooting

bricerisingalgorand: Ah you're right. Do you think it might be useful to be able to use a certain version rather than always running update? That might be convenient for testing/troubleshooting certain versions we've published

cicd/docker/algod/setup.py
bricerisingalgorand: That's kind of the rationale that led me to split it up this way. If we want to be able to set ports/tokens at any point, we would need a script that edits the respective config files to make that happen. Is there another way we could do that?

cicd/docker/indexer/Dockerfile
bricerisingalgorand: oh thanks, i'll fix that

docker-compose.yml
bricerisingalgorand: To do that you just need to run this command

```
mule -f cicd/mule.yaml build-local-images
```

cicd/docker/algod/entrypoint.sh
bricerisingalgorand: Cool, i'll make that change. Thanks!

bricerisingalgorand:
> Please don't move the feature files. We finally got things to the point where they don't break every time we update this repo, and that change will definitely break all of the SDKs when we make it.

Does it make more sense for them to be in their current folder? I moved them so that they can be dropped more conveniently into the sdk projects. To stay where they are wouldn't be too much of a problem though. 

> I'm not sure I understand the benefit of storing this in docker-hub. By definition, after 24h the containers running the "nightly" channel are no longer useful.

Storing these images in dockerhub is where a lot of the speed improvement in this process comes from. Also if we'd prefer to we can also run the update script on container startup in order to avoid storing every nightly build we make, since that still takes less time than rebuilding the images. Dockerhub is free storage for us though, so we don't really lose anything by uploading them right?

cicd/docker/algod/setup.py
winder: A certain version of go-algorand? Maybe, I can imagine if there was a regression caused by an algod change, running the SDKs against an older version would verify that algod caused the regression.

cicd/docker/algod/setup.py
winder: I think the docker way to do this would be to hard code the port in the container, and map it to a different host port when the container starts.

To update the tokens you have to update the token files then restart algod/kmd.

PR #87, opened 2020-06-01
fix old tests

winder:
The Java SDK compares all fields of the JSON files, so adding the unsupported field to the existing file broke that test.


features/unit/v2indexerclient_responsejsons/searchForAccounts_1.json
```diff
@@ -1,4 +1,3 @@
-http://localhost:8980/accounts?currency-less-than=10000000000&round=5975733&asset-id=979&limit=2&currency-greater-than=100000000&next=AABKF2BFMRR2OFS5HIIULF5CISPKFFV5RYFM2YTPCRWTBKPWKTOKWTSOME
```
winder, line 1: I'm not sure what this was all about, but we didn't seem to be using it anywhere so I'm guessing it was leftover from something we didn't need.

PR #88, opened 2020-06-01
Application Transaction Unit Test Goldens

winder:
This has some scripts that generated most of the feature file, and a cleaned up feature file for the new transaction builder unit tests.

It also has a start to the paths test for the client. This should be finished in another PR so that the transaction goldens can be merged.


features/unit/transactions.feature
```diff
@@ -0,0 +1,31 @@
+@unit.transactions
```
winder, line 1: Here is the feature file you should review, most of the other files are related to the scripts and can be ignored unless you're curious.

features/unit/v2indexerclient_paths.feature
```diff
@@ -165,3 +165,23 @@ Feature: Indexer Client v2 Paths
     Examples:
       |path            | account                                                    | notePrefixB64 | txType | sigType | txid                                                 | round| minRound | maxRound | limit | beforeTime              | afterTime               | currencyGreaterThan | currencyLessThan | index | addressRole | excludeCloseTo | rekeyTo |
       |/v2/transactions?rekey-to=true   |  |   |    |      |  | 0  |   0     | 0       | 0    | |  | 0                  | 0         | 0   |       |            |                                                                                            true|
+
+  @unit.applications
```
winder, line 169: Also the new path tests

utils/create_transactions.sh
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
+#
+# Create cucumber example tables for app transactions
+#
+# Must have an applications-enabled version of goal on the path, and any running node.
+
+
+# this is mandatory, 'goal app' currently only supports looking up fee/first-valid/last-valid
+DATA_DIR="/home/will/algorand/networks/applications-test/Node"
```
EvanJRichard, line 9: will need to be updated before merge, right?

utils/create_transactions.sh
```diff
@@ -0,0 +1,135 @@
...
+ACCOUNT_1=BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4
+ACCOUNT_MNUMONIC="awful drop leaf tennis indoor begin mandate discover uncle seven only coil atom any hospital uncover make any climb actor armed measure need above hundred"
```
EvanJRichard, line 11: nit: `mnemonic`

utils/create_transactions.sh
winder: This is just used as a utility to generate the feature file table, so it's mostly for reference in case something similar is useful in the future.

utils/create_transactions.sh
winder: fixed

utils/programs/asa_approve.teal
```diff
@@ -0,0 +1,531 @@
+txn NumAppArgs
```
algonautshant, line 1: Can you please add some comments on what this script is doing? 
An overview similar to utils/programs/globcheck.teal  with be super helpful here.

utils/programs/assetround.teal
```diff
@@ -0,0 +1,120 @@
+//
```
algonautshant, line 1: An overview similar to utils/programs/globcheck.teal  with be super helpful here.

utils/programs/asa_clear.teal
```diff
@@ -0,0 +1,9 @@
+byte base64 Ymw=
```
algonautshant, line 1: An overview similar to utils/programs/globcheck.teal  with be super helpful here.

utils/create_transactions.sh
algonautshant: This will be of great value to have for reuse, reference, or learning. 
Please do one more step of removing local reference (e.g. /home/will/...) and make it parametrized.

PR #90, opened 2020-06-09
Allow building test environment from source

winder:
Add several new options to `up.sh`. By default it behaves the same as before, starting a test environment using the `nightly` release channel. There is a new option to choose a dockerfile for algod which builds from source. There are also a couple other convenience options that I found useful while debugging the containers, or that we've been wanting during development.

There are some new ARGs in the Dockerfile's which can be configured in `docker-compose.yml`. I'm thinking that they could be exposed as `up.sh` parameters if they end up being useful enough. I wanted to configure the ARGs with a file, but it seems that docker only supports `env_file`. There is no `arg_file` that I could find.

These changes were heavily inspired by #85 and the TICK Sandbox.


docker/algod/source/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Basic dependencies.
+ENV HOME /opt
+RUN apt-get update && apt-get install -y apt-utils curl git git-core bsdmainutils python3
+
+# Copy everything into the container.
+COPY . /tmp
+
+# Install algod binaries.
+RUN git clone --single-branch --branch "${BRANCH}" "${URL}" && \
```
bricerisingalgorand, line 19: This isn't a functional issue, but it may be better to have this git clone be done by the up script, then use these docker steps to take advantage of docker caching better

```
COPY go-algorand/scripts/configure_dev.sh /tmp/go-algorand/scripts/configure_dev.sh
COPY go-algorand/scripts/configure_dev-deps.sh /tmp/go-algorand/scripts/configure_dev-deps.sh
RUN cd /tmp/go-algorand && \
         ./scripts/configure_dev.sh
COPY go-algorand /tmp/go-algorand
RUN make install
```

The same goes for the COPY . /tmp part too actually. When you add every file like that in the beginning, if any single file changes the entire docker build cache is reset. To take better advantage of it you can add individual files for processes that take a long time and then throw everything you need in at the end.

docker/algod/source/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM golang:1.13.12
```
bricerisingalgorand, line 1: You could actually utilize the other algod docker image in this one to make things a little faster using multi-stage builds like this:
```
FROM golang:1.13.12 as builder

RUN #Build binaries

FROM algod:latest
COPY --from=builder /algod/bin /algod/bin
```

Doing it this would reduce the amount of duplicated code to keep up with and make recurring builds a bit faster. It would requiring ensuring the algod image is available prior to building this one though.

docker-compose.yml
```diff
@@ -6,7 +6,14 @@ services:
     container_name: sdk-harness-algod
     build:
       context: .
-      dockerfile: ./docker/algod/Dockerfile
+      dockerfile: "./docker/algod/${TYPE}/Dockerfile"
```
bricerisingalgorand, line 9: You could use this to set a default for type

```
${TYPE:-channel}
```

docker/algod/source/Dockerfile
winder: I don't think caching helps much for this environment because we're building from source, and the source is constantly changing.

The only really cachable bit is libsodium, and in theory it could be cached by building master, then adding a different remote / checking out the target branch. I think that would add too much complexity for the speedup we'd see.

docker/algod/source/Dockerfile
winder: Cool, I didn't know you could combine multiple dockerfiles.

Just so I'm clear, this would require 3 dockerfiles? There would be separate dockerfiles to install from source, or from a channel, and then a 3rd to setup / configure the network?

docker-compose.yml
winder: Nice, `docker-compose up` this will let `docker-compose up` with no scripts to work again.

docker/algod/source/Dockerfile
bricerisingalgorand: So this isn't actually combining dockerfiles. What this is doing is executing two docker build stages. First it uses the golang base image to build algorand binaries, then in the second stage it copies the binaries that were built onto the already built algod image. Multi-stage docker builds produce images for each stage, but only the last image is tagged. So with this example we would have one docker image which is the algod image with it's binaries overwritten. So it's basically taking the docker image you've already made and overwriting the binaries with ones built from source. It would be two dockerfiles though. One for this process, and another for the algod image you've already written. Does that make sense?

Here's a doc that might help explain it
https://docs.docker.com/develop/develop-images/multistage-build/

docker/algod/source/Dockerfile
bricerisingalgorand: The docker build process actually accounts for that though. The hash computed for these copy steps is computed based on the contents that are copied over. In this case if `configure_dev.sh` or `configure_dev-deps.sh` change, the cache will be invalidated and and subsequent steps will be run, but if they don't the cached execution of those scripts can be used. This way you can save time by only running those scripts on changes. I'm with you though that taking advantage of this may be more work than it's worth for now, since our folder structure isn't really set up well for it.

winder: The source only exists in the container, so `git clone` needs to run in order to update these files

docker/algod/source/Dockerfile
winder: Thanks for the explanation, that does make sense. I actually did a crude version of this in the past by copying a locally build `algod` binary into the container.

I don't see how this is faster though, either way the private network needs to be created using the new binaries.

PR #93, opened 2020-06-13
Add applications integration test.

winder:
This is a scary one, but there are only 6 unique steps.

In order to use this, you need to modify `docker-compose.yml`:
```
        URL: "https://github.com/justicz/go-algorand"
        BRANCH: "maxj/applications"
```

And start the test environment with `up.sh --type source`

### DIFF
--- a/.up-env
+++ b/.up-env
@@ -1,0 +1,8 @@
+export TYPE="source"
+
+# Used when TYPE=channel
+export CHANNEL="nightly"
+
+# Used when TYPE=source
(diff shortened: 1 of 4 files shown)


winder:
@EvanJRichard things in quotes and numbers are parameters, even if you don't use the example table. This is the java implementation of that one:
```
@Given("I create and fund a transient new account with {long} microalgos.")
public void createAndFundTransientAccount(Long amount) throws Exception {
```

PR #94, opened 2020-06-29
Teal v2 tests for compile and dryrun

algorandskiy:


scripts/local/up.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
winder, line 1: please don't commit this file, the docker environment should be used to make sure testing is consistent across all of our environments.

features/integration/compile.feature
```diff
@@ -0,0 +1,13 @@
+@compile
+Feature: Compile
+  Background:
+    Given an algod v2 client
+
+  Scenario Outline: Compile programs
```
winder, line 6: These could probably be rewritten slightly so that the compile step only needs to be implemented once. Maybe something like this (4 steps instead of 6, per SDK)

```
When I compile a teal program <program>
Then it is compiled with <status> and <result> and <hash>

...

Given I compile a teal program <program>
When I dry run it
Then I get execution result <result>
```

scripts/local/up.sh
algorandskiy: I guess docker should be used in CI but not in local development. does it hurt to have it there?

features/integration/compile.feature
algorandskiy: In this case one moving part is added to dryrun test. My intention was in testing one REST API endpoint in these scenario. I plan to add couple more tests and have you idea in mind.

algorandskiy: Rewrote to
```
    When I dryrun a <kind> program <program>
    Then I get execution result <result>
```
Dryrun can accept both source and binary so `I compile a teal program` is exceed step.

scripts/local/up.sh
winder: The docker environment is configured so that you can run your tests locally against the containerized test environment. If the supported scripts aren't working we should fix them instead of creating additional scripts which aren't supported.

features/unit/v2algodclient_responses.feature
```diff
@@ -91,3 +91,11 @@ Feature: Algod REST API v2 Responses
     Examples:
       |jsonfiles           |directory                   |err|roundNum|
       |suggestedParams.json| v2algodclient_responsejsons||6222155|
+
```
winder, line 94: Add `@unit.dryrun` here, then I think we can merge this in

algorandskiy: done

PR #95, opened 2020-07-01
Add scripts for local algod/kmd startup

algorandskiy:
Running docker locally not great for development:
* hard to debug
* might need different environment rather than docker creates
* building algod twice (locally to check changes, and the same in docker is not time efficient)

Scripts `up.sh` and `down.sh` for local development eliminate above problems.

### DIFF
--- a/scripts/local/down.sh
+++ b/scripts/local/down.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+TEMP=$(dirname "$(mktemp -u)")
+NETWORK_NAME="sdknet"
+NETWORK_DIR="/$TEMP/$NETWORK_NAME/"
+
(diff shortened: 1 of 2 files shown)


algorandskiy:
> hard to debug

```bash
less test/docker/Dockerfile

FROM golang:1.13-stretch

# Copy SDK code into the container
RUN mkdir -p $HOME/go-algorand-sdk
COPY . $HOME/go-algorand-sdk
WORKDIR $HOME/go-algorand-sdk

# Run integration tests
CMD ["/bin/bash", "-c", "make unit && make integration"]
```

How would I break in IDE if docker not only runs algod/indexer but also runs tests?

winder:
When you're developing the SDK you would run cucumber from your IDE instead of with that Dockerfile

PR #96, opened 2020-07-02
Applications REST Client Response Tests

winder:
The response files look strange because they were generated from the spec. We can improve them in the future, but for now they will hopefully do the trick.

Rather than having many "I make any ____ call" functions, I have a single function which accepts a client/endpoint combination and we can fill them in.

Last but not least, instead of cherry picking a few fields to compare there is a step called "the parsed response should equal the mock response.". The intention is that the SDK response object can be serialized back into JSON, and the two JSON files can be compared against each other.


features/unit/responses.feature
```diff
@@ -0,0 +1,45 @@
...
+      | indexer_applications_TransactionsResponse_0.json        | 200    | indexer | searchForTransactions     |
+      | indexer_applications_ErrorResponse_0.json               | 500    | indexer | any                       |
+      | algod_applications_NodeStatusResponse_0.json            | 200    | algod   | GetStatus                 |
+      | algod_applications_NodeStatusResponse_0.json            | 200    | algod   | WaitForBlock              |
+      | algod_applications_PostCompileResponse_0.json           | 200    | algod   | TealCompile               |
+      | algod_applications_PostTransactionsResponse_0.json      | 200    | algod   | RawTransaction            |
+      | algod_applications_SupplyResponse_0.json                | 200    | algod   | GetSupply                 |
+      | algod_applications_TransactionParametersResponse_0.json | 200    | algod   | TransactionParams         |
+      | algod_applications_ErrorResponse_0.json                 | 500    | algod   | any                       |
+      # These ones failed to generate, figure out why in the responses generator.
```
winder, line 29: The response generator failed to create a few response objects... I'll get those added.

PR #105, opened 2020-07-20
Update applications test with a verification step.

winder:
This test requires adding one more step in addition to the `@application` tests. There is also a new examples table so that the localstate and globalstate are both tested.

Example implementation:
https://github.com/algorand/java-algorand-sdk/pull/182


features/integration/applications.feature
```diff
@@ -1,12 +1,13 @@
...
 Feature: Applications
    Background:
       Given an algod client
       And a kmd client
       And wallet information
       And an algod v2 client connected to "localhost" port 60000 with token "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
 
+   @applications
    Scenario: Use every applications feature in one scenario, do it!
+      # TODO: This test is a subset of @applications.verify - once everyone implements that one, get rid of this one.
```
algonautshant, line 10: Why not just remove it, and let the tests fail. and we'll update and fix it.

winder: I thought about it. Maybe it's fine to do that, but I'm trying to avoid breaking things that are currently working for now to avoid distracting people with false negatives. For example you're working on the response tests now, if I didn't implement it like this then your tests would start failing once this gets merged in.

PR #120, opened 2020-09-27
fix some errors

RileyGe:


winder:
Hi @RileyGe, what error is this fixing?

RileyGe:
@winder I used the feature in SpecFlow(dotnet) and found two errors saying "Feature file already contains a scenario with name 'XXXX'". 
The first error, we have two same scenarios "/transactions?max-round=<max-round>" in file features\unit\v2algodclient_paths.feature, so I deleted one of the scenarios.
The second error, we have two scenario outline: Pending Transaction Information, I renamed one of them to Pending Transaction Information2.

PR #122, opened 2020-10-23
Update indexer response for nonzero rewards

### LINKED ISSUE #121: Account rewards are expected to be 0
Yesterday the Javascript SDK started failing an integration test: https://travis-ci.com/github/algorand/js-algorand-sdk/jobs/403990024

The only difference between the actual and expected JSON is that some accounts have a nonzero `"reward"`, so I believe this is because https://github.com/algorand/indexer/pull/221 was merged.

I was going to make a PR that hardcodes the values from the actual response above, but since rewards are time dependent I wasn't sure if that would be too brittle.

jasonpaulos:
Regenerated indexer responses after running integration tests.

Closes #121.


winder (approved):
LGTM

PR #136, opened 2021-06-16
add test for checking the fee field is omitted when 0

shiqizng:


shiqizng:
I need to make some updates. will reopen it later.

PR #137, opened 2021-06-16
Omit fee field when it is 0

shiqizng:
Added unit tests to check that fee field is omitted when its value is 0.


jdtzmn (approved):
I don't have much experience with this repo, so I would hold off on merging until Jason can review this. That being said, I'd assume there should be a `Then` keyword somewhere in the file.

features/unit/feetest.feature
```diff
@@ -0,0 +1,23 @@
+@unit.feetest
+@unit
+Feature: Transaction fee test
+  Background:
+    Given a signing account with address "BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4" and mnemonic "awful drop leaf tennis indoor begin mandate discover uncle seven only coil atom any hospital uncover make any climb actor armed measure need above hundred"
+
+  Scenario Outline: Fee Test
+    When I build an application transaction with operation "<operation>", application-id <application-id>, sender "<sender>", approval-program "<approval-prog-file>", clear-program "<clear-prog-file>", global-bytes <global-bytes>, global-ints <global-ints>, local-bytes <local-bytes>, local-ints <local-ints>, app-args "<app-args>", foreign-apps "<foreign-apps>", foreign-assets "<foreign-assets>", app-accounts "<app-accounts>", fee <fee>, first-valid <first-valid>, last-valid <last-valid>, genesis-hash "<genesis-hash>", extra-pages <extra-pages>
+    And sign the transaction
+    And fee field is in txn
```
jasonpaulos, line 10: I agree with @jdtzmn, this op should be a `Then` instead of an `And`.
```suggestion
    Then fee field is in txn
```

jasonpaulos:
These look like good tests for app calls, but could you also add a scenario for a basic payment transaction?

features/unit/feetest.feature
```diff
@@ -0,0 +1,42 @@
...
+
+    Examples:
+    | fee | fv    | lv    | gh                                           | to                                                         | close                                                      | amt  | gen          | note         | mn                                                                                                                                                                   | addresses                                                                                                                                                                        | golden                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
```
jasonpaulos, line 41: Nit: I don't think `golden` is used here or in the previous scenario -- you could remove it.

jasonpaulos (approved):
One insignificant comment, but this looks good to merge.

PR #139, opened 2021-08-13
Update indexer dockerfile to build with new go-algorand dependency

bricerisingalgorand:
Indexer now depends on go-algorand, so our indexer docker image needs to have some go-algorand dependencies in it.

### DIFF
--- a/docker/indexer/Dockerfile
+++ b/docker/indexer/Dockerfile
@@ -6,7 +6,8 @@ ARG SHA=""
 
 RUN echo "Installing from source. ${URL} -- ${BRANCH} -- ${SHA}"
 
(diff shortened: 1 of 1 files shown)


algochoi (approved):
Thanks for this change, hope the build errors go away now

bricerisingalgorand:
Merging this now to unblock @algochoi , we can change if we need to when you get back @winder . We may want to switch the indexer image to ubuntu so that we can use glibc instead of musl

winder:
@bricerisingalgorand thanks for jumping on this. Looks good to me (especially if the changes worked!).

I suppose this means we have some broken sandbox configurations as well?

PR #141, opened 2021-08-31
stop testing deprecated contract template feature

brianolson:
Remove this first so that removing the feature from SDKs will cause no failures.
Part of https://github.com/algorand/go-algorand-internal/issues/1415


tzaffi:
Should be able to close PR and delete branch as template features has already been deleted.

tzaffi:
Closing because already addressed by #156 and the file no longer exists in repo's HEAD

PR #148, opened 2021-11-05
ABI Interaction integration tests with app calls

### LINKED ISSUE #146: Add tests for ABI interaction - AtomicTransactionComposer
Test support for SDKs invoking on-chain contracts with the ABI. Specifically, test the implementation of the AtomicTransactionComposer (the second feature) from [this design document](https://gist.github.com/jasonpaulos/a810abe7e86d43840d14445718565a9a) for each SDK.

This likely requires a new integration test feature file.

algochoi:
This PR introduces some Cucumber integration tests for interacting with ARC-0004 compliant apps using the `AtomicTransactionComposer` in the SDKs. 

PyTeal contract: [gist](https://gist.github.com/algochoi/fe831538208a616aa085f34061f303f1)

This change will likely require the SDKs to parse a new app-arg prefix, `hex` that will convert a hex value into bytes, e.g. `bytes.fromhex(some_hex_value)`. The return value is also logged in Base64, so it will need to be converted as well, e.g.  `base64.b64decode(ret).hex()`.

Closes #146 

Todo:
- [x] Standardize how return logs are handled (Hash the word `return`)


jasonpaulos (commented):
This appears to test `AtomicTransactionComposer` in a more indirect way than what I was expecting. I was imagining we would have new steps for each action of the composer and mix these in with some of the existing transaction construction steps. Something like:

```
Given a new atomic transaction composer
And suggested transaction parameters fee <fee>, flat-fee "<flat-fee>", first-valid <first-valid>, last-valid <last-valid>, genesis-hash "<genesis-hash>", genesis-id "<genesis-id>"
When I build a transaction ...
And I make a transaction signer for account mnemonic "<mnemonic>"
And I add the transaction and signer to the composer
And I gather signatures with the composer
Then the signatures should equal "<expected>"
```

Though admittedly we only have general txn construction steps for app and keyreg txns at the moment. If needed I can help flesh these out for the other txn types.

I may have lead you astray by suggesting we reuse steps as much as possible. In general we should do that, but it doesn't appear to apply to this test very much.

algochoi:
I only added a `PaymentTxn` test case for now since the current tests only require payments -- I can create a ticket for creating the remaining transaction types in the cucumber tests.

features/integration/abi.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I clone the composer.
+    # Create a payment method call with an address argument, and add it to the composer
+    And I build a method with signature "<method-signature>".
+    And I add a method call with the signing account, the current application, suggested params, operation "call", current transaction signer, app-args "<app-args>".
+    # Build the group in the composer
+    And I build the transaction group with the composer. If there is an error it is "".
+    Then The composer should have a status of "BUILT".
+    And I gather signatures with the composer.
+    Then The composer should have a status of "SIGNED".
+    And I submit the current transaction group with the composer.
```
algochoi, line 36: I am not sure if the signed transactions can be checked here because there are some variables like `app-id` that may differ during the integration tests. I put a check for the signed payment transaction above.

jasonpaulos: Good point. What if we split this into an integration and unit test?

The integration test would create the app, use the composer to build a group which calls a method on the app, then execute the composer and verify the return value.

The unit test(s) would not actually send anything to the network, but use the composer to build groups and verify output of intermediate stages.

features/integration/abi.feature
```diff
@@ -0,0 +1,48 @@
+@abi
+Feature: ABI Interaction
+  Background:
+    Given an algod v2 client
+    And a signing account with address "BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4" and mnemonic "awful drop leaf tennis indoor begin mandate discover uncle seven only coil atom any hospital uncover make any climb actor armed measure need above hundred"
+
+  Scenario: AtomicTransactionComposer test
+    Given a new AtomicTransactionComposer
+    And suggested transaction parameters fee 1234, flat-fee "true", first-valid 9000, last-valid 9010, genesis-hash "Mf0h6zjkEIEZPtNM3zsrg+iHQFS0fZxhgr7w35I464M=", genesis-id "cucumbernet"
```
jasonpaulos, line 9: I realized this won't work on an integration test anyway, since it uses hardcoded values. Probably we'll need a step like `And suggested transaction parameters from the algod v2 client` to get the values from algod

features/integration/abi.feature
```diff
@@ -0,0 +1,80 @@
...
+    And I execute the current transaction group with the composer.
+    Then The composer should have a status of "COMMITTED".
+    And The app should have returned "<returns>" in the log.
+
+    Examples:
+      | method-signature         | app-args                                                  | returns |
+      | add(uint64,uint64)uint64 | [int:1,int:1]                                             | 151f7c750000000000000002 |
+      | empty()void              | []                                                        | 151f7c75 |
+
+  Scenario: AtomicTransactionComposer with transaction arguments test
```
algochoi, line 48: I created another test scenario for method calls that require transactions as method arguments. Another approach was to have some `transaction-arg` in the Examples, but I thought it would get ugly in the SDKs as they parse for regex and hardcode certain transactions for the method call.

algochoi (commented):
Split up the tests so the unit tests check for the signed transaction goldens and the integration tests check for the app return values.

One tricky thing is how we build up transaction and app arguments for the method calls, and I've just chosen to hard-code it on the Cucumber side rather than the SDK side. It would be nicer if the transactions can be "parameterized" better, but I'm not sure if there is a better way right now.

features/integration/abi.feature
```diff
@@ -0,0 +1,80 @@
...
+    And I remember the new application ID.
+    # Create a pay transaction, create a TransactionSigner, and add it to the composer
+    When I make a transaction signer for the signing account.
+    And I remember the transaction signer.
+    And I build a payment transaction with sender "BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4", receiver "BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4", amount 100001, close remainder to ""
+    And I create a transaction with signer with the current transaction.
+    And I remember the transaction with signer.
+    # Create a payment method call with an address argument, and add it to the composer
+    And I build a method with signature "<method-signature>".
+    And I prepare the method arguments with the current transaction with signer and the "<app-args>".
```
jasonpaulos, line 64: I like the steps for method app args, but I'm not a fan of the argument format. Prefixes like `int:` and `hex:` work for encoding values to bytes for normal app calls, but I don't think they work as well for ABI types. We could implement general JSON parsing and conversion like goal does in https://github.com/algorand/go-algorand/pull/3088, that's a lot of work just for testing.

What if instead the argument format was a comma-separated list of the encoded ABI values in base64? For example, the args 2 and 3 for the method `add(uint64,uint64)uint64` could be given as `AAAAAAAAAAI=,AAAAAAAAAAM=`. It's not the nicest to look at but it would be pretty simple to decode and process in this step, even for complicated ABI values.

features/unit/transactions.feature
```diff
@@ -35,6 +35,19 @@ Feature: Transaction encoding
...
+  Scenario Outline: Payment Transaction Tests
+    Given suggested transaction parameters fee <fee>, flat-fee "<flat-fee>", first-valid <first-valid>, last-valid <last-valid>, genesis-hash "<genesis-hash>", genesis-id "<genesis-id>"
+    When I build a payment transaction with sender "<sender>", receiver "<receiver>", amount <amount>, close remainder to "<close_remainder_to>"
+    And sign the transaction
+    Then the base64 encoded signed transaction should equal "<golden>"
+    And the decoded transaction should equal the original
+    
+    Examples:
+      | fee  | flat-fee | first-valid | last-valid | genesis-hash                                 | genesis-id  | sender                                                     | receiver                                                   | amount | close_remainder_to | golden |
+      | 1234 | true     | 9000        | 9010       | Mf0h6zjkEIEZPtNM3zsrg+iHQFS0fZxhgr7w35I464M= | cucumbernet | BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4 | BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4 | 100001 |                    | dH6P7cTDI91JxHqi5E5056ty30cCwqzJpdCMKHD6WZo4kPB5nGq3aNrc1KBLYfWjyw2GEshjOoY7nSmGDkJLBQ== |
```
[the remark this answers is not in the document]
algochoi: Good catch -- I will fix the goldens (msgpack encoding of the signed transaction) to reflect this in the next commit...

features/integration/abi.feature
algochoi: Sounds good to me 👍

algochoi: Actually, if I'm think about this correctly, method arguments are in their native types in the respective SDKs. 

```
type ABIValue = boolean | number | bigint | string | Uint8Array | ABIValue[];
type MethodArgument = ABIValue | TransactionWithSigner;
```

So when `addMethodCall()` for `add` is called on the composer, it's passed in `[1, 1]`, where `1` is an `int` or `number`, etc. I'm inclined to go back to the prefix format so it can just cast the value based on that, but interested to hear your thoughts.

algochoi: edit: I think the ABI encoded values makes sense

algochoi:
I updated the tests and tested them against the Python SDK. This commit contains the relevant diff for implementing the test steps in the Python SDK: https://github.com/algorand/py-algorand-sdk/pull/247/commits/e96b74049d4cdbec9c5efdfdd9ba4eaebfc8e09a

features/unit/atomic_transaction_composer.feature
```diff
@@ -0,0 +1,54 @@
...
+    And suggested transaction parameters fee 1234, flat-fee "true", first-valid 9000, last-valid 9010, genesis-hash "Mf0h6zjkEIEZPtNM3zsrg+iHQFS0fZxhgr7w35I464M=", genesis-id "cucumbernet"
+    # Save an application id so we can test method calls
+    And an application id <app-id>
+    # Create a pay transaction, create a TransactionSigner, and add it to the composer
+    When I make a transaction signer for the signing account.
+    And I build a payment transaction with sender "BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4", receiver "BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4", amount 100001, close remainder to ""
+    And I create a transaction with signer with the current transaction.
+    # Create a method call with an address argument, and add it to the composer
+    And I build a method with signature "<method-signature>".
+    And I prepare the method arguments with the current transaction with signer and the app args "<app-args>".
```
jasonpaulos, line 42: I think this could be more expressive, since currently we're limited to having exactly one transaction at index 0. What about something like this?

```
And I create a new method arguments array
And I append the current transaction with signer to the method arguments array
And I append the encoded arguments "<app-args>" to the method arguments array
And I add a method call with the signing account, the current application, suggested params, operation "call", current transaction signer, current method arguments.
```

By splitting up the method args construction into multiple steps, we can support many more cases, even ones with multiple transactions.

I also have a nit: could you change `operation "call"` to `on complete "noop"`? I think that makes more sense with how we're using app calls here

features/integration/abi.feature
```diff
@@ -0,0 +1,78 @@
...
+    Then The composer should have a status of "BUILT".
+    And I gather signatures with the composer.
+    Then The composer should have a status of "SIGNED".
+    # Execute the group and check that the log contains the correct return value
+    And I execute the current transaction group with the composer.
+    Then The composer should have a status of "COMMITTED".
+    # Check that the log contains the expected value.
+    # Note that the <returns> is the exact log output, which is the base64 encoding
+    # of the value, with the first four bytes being the hash of "return"
+    And The app should have returned "<returns>" in the log.
```
jasonpaulos, line 43: nit: I think `And the app should have returned "<returns>"` would be slightly better, where `<returns>` is the ABI encoded expected return value.

I don't see an advantage to including the 4 byte return prefix here, since this step isn't meant to comb through all the logged values, it's only meant to validate what the composer says is the return value.

jasonpaulos (approved):
Looks good!

PR #149, opened 2021-11-08
SDK Tests for ABI JSON encoding/decoding

algoidurovic:
Tested with go-algorand-sdk


features/unit/abijson.feature
```diff
@@ -0,0 +1,75 @@
...
+@unit
+Feature: AbiJson
+  Scenario Outline: Serialize Method object from sig into json 
+    When I create the Method object from "<methodsig>"
+    And I serialize the Method object into json
+    Then the produced json should equal "<jsonfile>" loaded from "<directory>"
+
+    Examples:
+      | methodsig                | jsonfile       | directory         |
+      | add(uint32,uint32)uint32 | addMethod.json | abi_responsejsons |
```
jasonpaulos, line 11: Can you add the following examples too?
* A method with `void` return type
* A method with no args
* A method with a tuple argument
* A method with a tuple return type

features/unit/abi_responsejsons/addMethod.json
```diff
@@ -0,0 +1 @@
+{"name":"add","args":[{"name":"","type":"uint32"},{"name":"","type":"uint32"}],"returns":{"type":"uint32"}}
```
jasonpaulos, line 1: I think the args should be `[{"type":"uint32"},{"type":"uint32"}]`, i.e. an empty name should be dropped.

Also applies to other JSON files

features/unit/abijson.feature
```diff
@@ -0,0 +1,75 @@
+@unit.abijson
+@unit
+Feature: AbiJson
+  Scenario Outline: Serialize Method object from sig into json 
+    When I create the Method object from "<methodsig>"
+    And I serialize the Method object into json
+    Then the produced json should equal "<jsonfile>" loaded from "<directory>"
```
jasonpaulos, line 7: I'd recommend also adding a final step to test decoding the response. Something like `And the deserialized Method object equals the original`, which would decode the JSON and compare it to the original Method object. 

Same suggestion for Interface and Contract.

features/unit/abijson.feature
```diff
@@ -0,0 +1,75 @@
...
+      | add   | first        | uint32       | second        | uint32        | uint32     | addMethodWithArgNames.json | abi_responsejsons |
+
+  Scenario Outline: Create and serialize Method object into json with description
+    When I create the Method object with "<name>" "<methoddesc>" "<firstargtype>" "<firstdesc>" "<secondargtype>" "<seconddesc>" "<returntype>"
+    And I serialize the Method object into json
+    Then the produced json should equal "<jsonfile>" loaded from "<directory>"
+
+    Examples:
+      | name  | methoddesc  | firstargtype | firstdesc   | secondargtype | seconddesc  | returntype | jsonfile                      | directory         |
+      | add   | description | uint32       | description | uint32        | description | uint32     | addMethodWithDescription.json | abi_responsejsons |
```
jasonpaulos, line 38: You could increase the strength of these tests by not using repeated values for the parameters. This specific test uses `uint32` for all types and `description` for all description, but a stronger test would vary these values to guard against a buggy implementation mixing them up.

jasonpaulos (commented):
Generally looks good, but I left a few suggestions to improve the tests.

Also, we generally follow the convention of putting the "argument name" into the cucumber test string too. This helps make step strings unique and avoids collision between them.

For example, `I create the Method object with "<name>" "<firstargtype>" "<secondargtype>" "<returntype>"` should probably be `I create the Method object with name "<name>" first arg type "<firstargtype>" second arg type "<secondargtype>" and return type "<returntype>"`.

Otherwise the step looks like `I create the Method object with ? ? ? ?` to cucumber, which is not the most informative.

jasonpaulos (approved):
Looks good!

PR #151, opened 2021-12-07
fmt documents

barnjamin:
Just used the formatter to format the tables, should only be whitespace changes with the exception of renaming "Scenarios" to "Examples"


features/integration/dryrun.feature
```diff
@@ -6,7 +6,7 @@ Feature: Dryrun
   Scenario Outline: Dryrun execution with binary and source programs
     When I dryrun a <kind> program <program>
     Then I get execution result <result>
-    Scenarios:
+      Scenarios:
```
barnjamin, line 9: This spacing is unexpected, I need to double check the syntax

winder (changes requested):
LGTM. Could you add a comment to the README that describes how you ran the formatter?

barnjamin:
@winder I just used whatever is built into vscode, though I should probably write a small script to pull down something like https://github.com/antham/ghokin and format it

winder (approved):

PR #156, opened 2021-12-31
Testing Contract to Contract Functionality on the SDK's 

tzaffi:
## New SDK Functionality to Be Tested
* `applicationAddress(appId)` - a way of converting an application's index to its account / address on the network
* Ability to handle and parse nested inner transactions including their group id and their EvalDelta's. In particular, it is recommended that the`AtomicTransactionComposer` persist in memory responses from the `PendingTransactionInfo/{TXID}` endpoint for later transaction tree analysis.

## Example SDK PR that Implements
[Python #268](https://github.com/algorand/py-algorand-sdk/pull/268)

## Linked Future Issues in Each SDK
* https://github.com/algorand/py-algorand-sdk/issues/275
* https://github.com/algorand/java-algorand-sdk/issues/289
* https://github.com/algorand/js-algorand-sdk/issues/494
* https://github.com/algorand/go-algorand-sdk/issues/280

## Summary of Changes

* `.up-env` pointing  test harness to the c2c branch [feature/contract-to-contract](https://github.com/algorand/go-algorand/tree/feature/contract-to-contract). _This is temporary to help developing other SDK's, and should not be merged as is_
* Linting and updating the README
* `docker/algod/source/Dockerfile` - upgrading the test-harness source channel to go 1.6.11
* integration test updates:
  * `applications.feature` - check for the ability to calculate an app's account address hash
  * `c2c.feature` - new test for C2C (see the cucumber section below for more details)
    * **HEADS UP go SDK DEVELOPER**: this test uses Cucumber's "*" step prefix, which is a synonym of **And**. It is an open question whether go's cucumber implementation can handle it
  * `compile.feature` - test the ability for the cucumber test to compile `*.teal` files and run a test without accompanying bytecode
* Removing the `template.feature` integration test (merged in #141 ) (see **Cucumber Steps that Have Been Removed. SDK's should do the same.** below)
* New C2C TEAL programs used in the tests: `slot_machine.teal` &rarr; `random_byte.teal` &rarr; `fake_random.teal`

### Linked Issue

#154 

## Cucumber Steps that Need Supporting or Updating

* `integrations/applications.feature` **NEW STEP**:
  * **Then** _I get the account address for the current application and see that it matches the app id's hash_
* `integrations/compile.feature` **NEW STEP**:
  * **Then** _base64 decoding the response is the same as the binary <program>_
* `integration/c2c.feature`:
  * **NEW STEPS**:
    * **Given** _I fund the current application's address with {fund_amount} microalgos._ 
      * Allow funding of apps
    * **Given** _I add the nonce "{nonce}"_
      * add the provided `nonce` to the context to be used by _I add a nonced method..._
  *  **Given** * _I reset the array of application IDs to remember._
    * Was needed because unrelated tests were failing on the go SDK due to the slice of app-ids persisting between tests.
  * **When** _I add a nonced method call with the {accountType} account, the current application, suggested params, on complete "{operation}", current transaction signer, current method arguments._
      * similar to "**When** _I add a method call..._" but adding the ability to uniquify transactions so can make identical method calls twice in the same atomic transaction. This requires implementing the _I add a nonce.._ step above, and the test should pick up the added nonce and use it to uniquify the method call
    * **Then** _The app should have returned ABI types "{abiTypes:MaybeString}"._
      * Similar to _The app should have returned ..._ but only checks the types. This is useful for probabilistic functions.
    * **REMOVED** ~**Then**** _I can dig into the resulting atomic transaction execution tree with path "{path}"_~
      * Allows checking to see that transaction responses have been _persisted_ and that their inner transactions  exist to arbitrary depth.
    * **Then** _I dig into the paths "{paths}" of the resulting atomic transaction tree I see group ids and they are all the same_
      * Allows checking that inner transactions have the `grp` field and that those within the same group have the same group id
    * **Then** _I can retrieve all inner transactions that were called from the atomic transaction with call graph "{callGraph}"_
      * add functionality to the test class so that it can generate an  _ABI transactions trace-graph_  that looks something like [this gist](https://gist.github.com/tzaffi/6166662066920ca321b598bc56389c05) and use it to verify that responses are as expected
    * REMOVED ~**Then** _I can dig the {i}th atomic result with path "{path}" and see the eval delta field "{field}"_
      * Check eval delta fields coming from inner transactions at arbitrary depth~
  * **PRE-EXISTING STEPS** with **updated functionality**:
    * **When** _I build an application transaction with the..._ (there are 3 variants)
      * add functionality to compile teal directly. I.e., build the app from `*.teal` and not `*.teal.tok`.
    * **Given** _I remember the new application ID._
      * In addition to remembering the new app id, this step should also append to a list of app ids that have been created and subsequently add into the ABI method arguments using new parsing syntax
    * **When** _I add a method call with the ...._ (there are 3 variants)
      * add the ability to handle arguments like `"ctxAppIdx:0,ctxAppIdx:1"` indicating to look at the remembered list of app ids. Note, these arguments are passed in another step **When** _I append the encoded arguments..._

## New Cucumber Steps that are Specific Only to the new TEAL Programs
* **Then** _The {resultIndex}th atomic result for randomInt({input}) proves correct_
* **Then** _The {resultIndex}th atomic result for randElement("{input}") proves correct_
* **Then** _The {resultIindex}th atomic result for "spin()" satisfies the regex "{regex}"_

## Cucumber Steps that Have Been Removed. SDK's should do the same.
These were only supported in the deleted file `integration/template.feature`. It is recommended that separate issues in each SDK be created for these step removals.
* **Given** _a split contract with ratio <ratn> to <ratd> and minimum payment <min_pay>_
* **When** _I send the split transactions_
* **Given** _an HTLC contract with hash preimage "<preimage>"_
* **When** _I fund the contract account_
* **When** _I claim the algos_
* **Given** _a periodic payment contract with withdrawing window <wd_window> and period <period>_
* **When** _I claim the periodic payment_
* **Given** _a limit order contract with parameters <ratn> <ratd> <min_trade>_
* **When** _I swap assets for algos_
* **_Given_** a dynamic fee contract with amount <amt>
* **_Given_** I send the dynamic fee transactions
* **Given** _Contract test fixture_


features/integration/compile.feature
```diff
@@ -10,3 +10,26 @@ Feature: Compile
       | program                 | status | result     | hash                                                         |
       | "programs/one.teal"     | 200    | "AiABASI=" | "YOE6C22GHCTKAN3HU4SE5PGIPN5UKXAJTXCQUPJ3KKF5HOAH646MKKCPDA" |
       | "programs/invalid.teal" | 400    | ""         | ""                                                           |
+
+
```
tzaffi, line 14: In theory (and in my own practice in the Python SDK) it is possible to omit the "*.tok" TEAL binary, and have `algod` do the compilation for you. This makes it a bit easier to write/tweak TEAL to make it feat the integration tests. But unveiling this ability would require some additional support in the other SDK's. So it's worth further discussion.

features/integration/template.feature
```diff
@@ -1,70 +0,0 @@
-@templates
```
tzaffi, line 1: Incorporating @brianolson 's PR (#141 ) to remove this deprecated test.

.up-env
```diff
@@ -1,8 +1,8 @@
-export TYPE="channel"
```
tzaffi, line 1: revert these changes when `feature/contract-to-contract` is fully merged

README.md
```diff
@@ -1,6 +1,10 @@
 # algorand-sdk-testing
+
 Testing files for Algorand SDKs
```
tzaffi, line 3: [Markdownlint](https://github.com/DavidAnson/markdownlint/tree/v0.24.0) which I have as a VS-code extension, re-spaced the README. Are we okay with these changes?

README.md
```diff
@@ -11,42 +15,60 @@ We have different feature files for unit and integration tests. The unit tests s
 
 # Test Descriptions
 
-| tag                             | description |
```
tzaffi, line 14: Re-organizing the table:
* Separate into unit and integration tests
* Add missing tests
* Remove stale ones
* Alphabetize

docker/algod/source/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.13.12
+FROM golang:1.16.11
```
tzaffi, line 1: Upgrade to latest version. Was failing without due to stale SSL certs.

features/integration/applications.feature
```diff
@@ -21,8 +21,9 @@ Feature: Applications
       And I sign and submit the transaction, saving the txid. If there is an error it is "".
       And I wait for the transaction to be confirmed.
       And I remember the new application ID.
+      Then I get the application info for the current application, and its account matches the app id's hash
```
tzaffi, line 24: New cucumber test to check that can read an application account from the SDK's `application-info` method.

features/integration/abi.feature
```diff
@@ -28,7 +28,7 @@ Feature: ABI Interaction
     When I create the Method object from method signature "<method-signature>"
     And I create a new method arguments array.
     And I append the encoded arguments "<app-args>" to the method arguments array.
-    And I add a method call with the transient account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments.
+    And I add a hash-deterministic method call with the transient account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments.
```
tzaffi, line 31: All pre-existing method calls in cucumber tests are `hash-deterministic` but the C2C tests need a `hah-nondeterministic` variant. This will require refactoring the `I add method call...` steps (there are 3 variants)

docker/algod/source/Dockerfile
ahangsu: 👍

features/resources/programs/slot_machine.teal
```diff
@@ -0,0 +1,226 @@
...
+
+    byte "reel"
+    int 2
+    itob
+    concat
+    byte "@!------"
+    app_global_put
+
+    // P(!!!) = 1/144, P(@@@) = 1/72
+    // With respective prizes of 50 Algos & 40 Algos expected payoff is 0.903 Algos
```
ahangsu, line 225: nit: would be nice to add some notes for explanation:
```
Assuming that each face of a reel shows with equal probability, and the randomness of each reel is independent.
```

tzaffi: 👍

features/integration/applications.feature
```diff
@@ -21,8 +21,9 @@ Feature: Applications
       And I sign and submit the transaction, saving the txid. If there is an error it is "".
       And I wait for the transaction to be confirmed.
       And I remember the new application ID.
+      Then I get the account address for the current application and see that it matches the app id's hash
```
tzaffi, line 24: New cucumber test to check that can calculate an app's account/address from its appid.

features/integration/c2c.feature
```diff
@@ -0,0 +1,107 @@
...
+    And The app should have returned ABI types "(byte[3],byte[17],byte[17],byte[17]):(byte[3],byte[17],byte[17],byte[17])".
+    And I can dig into the resulting atomic transaction execution tree with path "0,0,0"
+    And I can dig into the resulting atomic transaction execution tree with path "0,2,0"
+    And I can dig into the resulting atomic transaction execution tree with path "1,2,0"
+    And I dig into the paths "0,0:0,1:0,2" of the resulting atomic transaction tree I see group ids and they are all the same
+    And I can retrieve all inner transactions that were called from the atomic transaction with call graph "[{'spin(application,application)(byte[3],byte[17],byte[17],byte[17])':[{'appl':'appl'},{'appl':'appl'},{'appl':'appl'}]},{'spin(application,application)(byte[3],byte[17],byte[17],byte[17])':[{'appl':'appl'},{'appl':'appl'},{'appl':'appl'}]}]".
+    And The 0th atomic result for "spin()" satisfies the regex "^[@!-][@!-][@!-]$"
+    And The 1th atomic result for "spin()" satisfies the regex "^[@!-][@!-][@!-]$"
+    And I can dig the 0th atomic result with path "inner-txns.0.inner-txns.0.global-state-delta.0.key" and see the eval delta field "Y291bnRlcg=="
```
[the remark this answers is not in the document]
tzaffi: I'm not sure how we would unify. Maybe something like:
```
... with the path "innerTxns,0,innerTxns,0,globalStateDelta,0,key"...
```
?

features/integration/c2c.feature
```diff
@@ -0,0 +1,107 @@
+@c2c
+Feature: Contract to Contract Interaction
+  Scenario:
```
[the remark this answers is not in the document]
tzaffi: 👍

features/integration/c2c.feature
```diff
@@ -0,0 +1,115 @@
...
+    * I create a new method arguments array.
+    * I append the encoded arguments "AAAAAAAAACo=" to the method arguments array.
+    * I add a nonced method call with the transient account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments.
+
+    # randInt(1337) -> (r, witnessString)
+    Given I add the nonce "Thing Two"
+    When I create the Method object from method signature "randInt(uint64)(uint64,byte[17])"
+    * I create a new method arguments array.
+    * I append the encoded arguments "AAAAAAAABTk=" to the method arguments array.
+    * I add a nonced method call with the transient account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments.
```
tzaffi, line 31: Now adding the `nonce` explicitly

README.md
```diff
@@ -11,42 +15,60 @@ We have different feature files for unit and integration tests. The unit tests s
...
+| @abi                   | Test the Application Binary Interface (ABI) with atomic txn composition and execution. |
+| @algod                 | General tests against algod REST endpoints.                                            |
+| @auction               | Encode and decode bids for an auction.                                                 |
+| @application.evaldelta | Test that eval delta fields are included in algod and indexer.                         |
+| @applications.verified | Submit all types of application transactions and verify account state.                 |
+| @assets                | Submit all types of asset transactions.                                                |
+| @c2c                   | Test Contract to Contract invocations and injestion.                                   |
+| @compile               | Test the algod compile endpoint.                                                       |
+| @dryrun                | Test the algod dryrun endpoint.                                                        |
+| @dryrun.testing        | Test the testing harness that relies on dryrun endpoint. Python only.                  |
```
winder, line 60: A section for the tests which aren't universally implemented would be good too. I believe `@application.evaldelta` is only implemented by the Java SDK.

README.md
winder: Looks good to me. If you want to avoid disabling rules you could fix the headers too.

features/integration/c2c.feature
```diff
@@ -0,0 +1,115 @@
+@c2c
+Feature: Contract to Contract Interaction
+  Scenario:
+    * an algod v2 client
+    * a kmd client
+    * wallet information
+    * suggested transaction parameters from the algod v2 client
+    * I create a new transient account and fund it with 100000000 microalgos.
+    * I make a transaction signer for the transient account.
```
winder, line 9: Have we used the `*` syntax before? The quality of the cucumber implementations varies quite a bit when it comes to implementing the spec. In particular, I would keep an eye out on go plugin. Other than that, nice trick!

features/integration/c2c.feature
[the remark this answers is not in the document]
winder: You could also define a file format for checking all results at once to the expected output file, maybe it would need some sort of placeholders for non-deterministic values?

features/integration/c2c.feature
ahangsu: 👍

winder (approved):
I didn't review the C2C code, but the test and cucumber updates look good.

Thinking out loud about how we use cucumber in general...

It seems we have a sort of "shadow SDK" implemented in the form of cucumber functions. I guess these map pretty closely to some of the new functionality? Figuring out the right granularity of abstraction was always a challenge for us.

It might be possible to simplify things by focusing on unit tests. i.e. transaction building and response parsing. Maybe still submit the txn to make sure that the network accepts it. Definitely not as robust of a test as you have here, but maybe it would make it easier for table-driven tests that exercise more inputs?

For responses, we have a "Response Test Generator" as part of the generator project. It's not the best tool we have, and maybe not directly applicable here, but worth keeping in mind when changing transactions: [See "responses" near this point of the README](https://github.com/algorand/generator#build-as-a-self-executing-jar)

features/integration/compile.feature
```diff
@@ -12,20 +12,9 @@ Feature: Compile
...
-
-
-  Scenario Outline: Compiled teals are as expected - skip the hash check
-    When I compile a teal program <teal>
-    Then base64 decoding the response <result> is the same as the binary <program>
+    Then base64 decoding the response is the same as the binary <program>
     Examples:
       | teal                            | program                             | result |
```
jasonpaulos, line 19: nit: result column is unused

jasonpaulos (commented):
Mostly looks good, just a nit

features/integration/compile.feature
tzaffi: 👍

features/integration/c2c.feature
```diff
@@ -0,0 +1,115 @@
...
+    # Atomic Transaction Execution and Analysis
+    Then I build the transaction group with the composer. If there is an error it is "".
+    And I gather signatures with the composer.
+    And I execute the current transaction group with the composer.
+    And The composer should have a status of "COMMITTED".
+    And The app should have returned ABI types "(byte[3],byte[17],byte[17],byte[17]):(byte[3],byte[17],byte[17],byte[17])".
+    And I can dig into the resulting atomic transaction execution tree with path "0,0,0"
+    And I can dig into the resulting atomic transaction execution tree with path "0,2,0"
+    And I can dig into the resulting atomic transaction execution tree with path "1,2,0"
+    And I dig into the paths "0,0:0,1:0,2" of the resulting atomic transaction tree I see group ids and they are all the same
```
algochoi, line 109: Can we merge this step with the three steps above? It seems like this step already checks if it can dig into that path implicitly

algochoi (approved):
Looks good to me, I really like the slot machine example 🤑  I think the steps look good right now, and we can change some of the details later on if necessary.

features/resources/programs/slot_machine.teal
```diff
@@ -0,0 +1,229 @@
...
+callsub extract_element_and_witness
+// >>> randByte2 <-- witness2 <-- randByte1 <-- witness1 <-- randByte0 <-- witness0
+
+dup2
+pushbytes "witness0"
+callsub debug
+pop
+pushbytes "randByte0"
+callsub debug
+pop
```
ahangsu, line 152: The debugging helper is really useful! I suppose it should log some message like
```
dbg [witness_i] --->
** witness_i content **
```
to log, and the whole block of code is to keep track of stuffs before ABI return, is that correct?

README.md
tzaffi: 👍

features/integration/c2c.feature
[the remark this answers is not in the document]
tzaffi: Will take feedback from others who are implementing this step to see if there's a better version

features/integration/c2c.feature
```diff
@@ -0,0 +1,115 @@
...
+    * I create a new method arguments array.
+    * I append the encoded arguments "AAdnb29kYnll,ctxAppIdx:0" to the method arguments array.
+    * I add a nonced method call with the transient account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments.
+
+    Then I build the transaction group with the composer. If there is an error it is "".
+
+    Then I gather signatures with the composer.
+    And I execute the current transaction group with the composer.
+    Then The composer should have a status of "COMMITTED".
+    And The app should have returned ABI types "(byte,byte[17]):(byte,byte[17])".
```
algochoi, line 71: Does the "ABI types" string need to be repeated?
```suggestion
    And The app should have returned ABI types "(byte,byte[17])".
```

README.md
tzaffi: 👍

features/integration/c2c.feature
tzaffi: Will make a note that go should watch out

features/integration/compile.feature
```diff
@@ -10,3 +10,13 @@ Feature: Compile
       | program                 | status | result     | hash                                                         |
       | "programs/one.teal"     | 200    | "AiABASI=" | "YOE6C22GHCTKAN3HU4SE5PGIPN5UKXAJTXCQUPJ3KKF5HOAH646MKKCPDA" |
       | "programs/invalid.teal" | 400    | ""         | ""                                                           |
+
+
+  Scenario Outline: Teals compile to their associated binary
+    When I compile a teal program <teal>
+    Then base64 decoding the response is the same as the binary <program>
```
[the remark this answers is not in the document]
tzaffi: It's actually base64 decoding the base64 encoded response. So I prefer to keep as is.

features/integration/c2c.feature
tzaffi: 👍 (removing lines 106-108)

features/resources/programs/slot_machine.teal
tzaffi: Yes, that's what it does... but you'll quickly run out of space if you put too many messages in. That's why I have a bunch of them commented out.

features/integration/c2c.feature
tzaffi: Yes, because there are 2 separate calls to `randElement()` inside of the atomic transactions, so their responses should be at corresponding indexes. I guess it's possible to simplify the test to only handle a call of the same method again and again, and assume that the responses all have the same type, but that would break the python test as currently written.

algochoi: Got it, makes sense

tzaffi:
My plan is to not merge this in until branch `go-algorand/feature/contract-to-contract` is merged. That's because I don't want to have the main branch of this repo point to a non-main branch of `go-algorand`.

features/integration/c2c.feature
```diff
@@ -0,0 +1,172 @@
...
+    * I create a new method arguments array.
+    * I append the encoded arguments "AAAAAAAABTk=" to the method arguments array.
+    * I add a nonced method call with the transient account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments.
+
+    Then I build the transaction group with the composer. If there is an error it is "".
+    Then I gather signatures with the composer.
+    And I execute the current transaction group with the composer.
+    Then The composer should have a status of "COMMITTED".
+    And The app should have returned ABI types "(uint64,byte[17]):(uint64,byte[17])".
+    And The 0th atomic result for randomInt(42) proves correct
```
algochoi, line 38: I think Cucumber interprets anything in parentheses in a special way (optional text), similar to curly braces: https://github.com/cucumber/cucumber-expressions#optional-text. Maybe we can wrap it around parentheses or wordify the argument instead of using naked parentheses.

tzaffi: ouch... will investigate. But my first reaction is "really cucumber?"

algochoi: In JS, you just have to escape the parentheses within the step:
```
'The {int}th atomic result for randomInt\\({int}) proves correct'
```
It looks like the other SDKs do not suffer from the same fate so this comment can be resolved as is 😄

docker-compose.yml
```diff
@@ -135,7 +135,7 @@ services:
     volumes:
       - ./docker/indexer/init-scripts:/docker-entrypoint-initdb.d
     ports:
-      - 5433:5432
+      - 65432:5432
```
tzaffi, line 138: to avoid clashing against other local sandboxes

jasonpaulos (approved):

features/integration/applications.feature
```diff
@@ -12,7 +12,8 @@ Feature: Applications
       Given I create a new transient account and fund it with 100000000 microalgos.
       # Application create with extra pages should succeed with expected message.
       And I build an application transaction with the transient account, the current application, suggested params, operation "create", approval-program "programs/big_app_program.teal.tok", clear-program "programs/big_app_program.teal.tok", global-bytes <global-bytes>, global-ints 0, local-bytes <local-bytes>, local-ints 0, app-args "", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 3
-      And I sign and submit the transaction, saving the txid. If there is an error it is "logic eval error: pc=704 dynamic cost budget exceeded, executing intc_1: remaining budget is 700 but program cost was 701".
+      And I sign and submit the transaction, saving the txid. If there is an error it is "logic eval error: pc=704 dynamic cost budget exceeded, executing intc_1: local program cost was 701. Details: pc=704, opcodes=intc_1 // 1".
```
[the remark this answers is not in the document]
ahangsu:
```suggestion
      And I sign and submit the transaction, saving the txid. If there is an error it is "logic eval error: pc=704 dynamic cost budget exceeded, executing intc_1: local program cost was 700. Details: pc=704, opcodes=intc_1 // 1".
```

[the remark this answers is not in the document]
tzaffi: 👍

PR #161, opened 2022-02-05
adding dryrun trace test and expected app trace

barnjamin:
Testing format of the dryrun trace printer being added to the SDKs

https://github.com/algorand/py-algorand-sdk/pull/283
https://github.com/algorand/go-algorand-sdk/pull/289
https://github.com/algorand/js-algorand-sdk/pull/514
https://github.com/algorand/java-algorand-sdk/pull/305


tzaffi (commented):
suggesting another test case

features/unit/dryrun_trace.feature
```diff
@@ -0,0 +1,13 @@
+@unit
+Feature: Dryrun Trace
+
+    @unit.dryrun.trace.application
+    Scenario Outline: Application Trace Tests
+        Given a dryrun response file "<dryrun-response-file>" and a transaction id "<txn-id>"
+        When I call app trace
+        Then the output should equal "<golden>"
```
tzaffi, line 8:
```suggestion
        Then calling app trace produces "<app-trace-file>"
```
I'm recommending this step consolidation because `the output should equal` step seems too generic. I would guess something like that already exists, in which case you can re-use that step.

Also I'm recommending renaming `golden` to `app-trace-file` because `golden` usually refers to an inline string, while `*-file` is more clearly referring to a file.

features/unit/dryrun_trace.feature
```diff
@@ -0,0 +1,13 @@
...
+
+        Examples:
+            | dryrun-response-file                                 | txn-id | golden                                          |
```
tzaffi, line 11:
```suggestion
            | dryrun-response-file                                 | txn-id | app-trace-file                                          |
```

See my previous comment

tzaffi (commented):
A couple of suggestions around naming conventions and removing a step that seems a bit redundant.

A possible future improvement -but not worth holding up this PR for- is to do one of the following (or even both):

* add the possibility of CSV output (with no truncation). So would add example files and a step such as:
  * `Then calling app trace CSV produces "<app-trace-csv-file>"`
* add the possibility of non-truncated output. So would add example files and a step such as:
  * `Then calling app trace no-truncation produces "<app-trace-notrunc-file>"`

tzaffi (approved):
LGTM


For future reference, recommend opening an issue for possible improvements cited above (CSV + no truncation).

Also new issue that cropped up: handling errors in the trace. EG:

```
348 |    43 |   25 | swap                   |           | [21, 2432902008176640000, 21]
349 |    45 |   26 | store 0                | 0->21     | [21, 2432902008176640000]
350 |    45 |   26 | * overflowed           |           | [21, 2432902008176640000]
```

Which we can get from a switch that looks like this:

    def line_or_err(i, ln):
        line = lines[ln - 1]
        err = trace[i].get("error")
        return err if err else line

PR #163, opened 2022-03-02
fixes, file renames

algonautshant:
- Replace `abc` with proper address.  `abc` will be rejected by the java sdk as invalid address.
- rename the new unlimited assets response file prefixes appropriately. 
- add newline at the end of the files


tolikzinovyev:
Thanks.

Why have `unlimited_aa_` in the file names? Their directory already has `unlimited_assets_` in it.
Do you have a corresponding PR for the generator?

algonautshant:
I am not particular about the naming. "applications" was wrong, so this is unlimited apps/assets (aa).
This is the generator update PR: https://github.com/algorand/generator/pull/26

tolikzinovyev:
I think we should just erase `unlimited_aa_`. It's not needed.

tolikzinovyev:
Did you modify the generator to add the newlines? I couldn't find that change. Also, I'm not sure newlines are really necessary.

algonautshant:
the newlines at the end of the files are not added by the generator.

PR #164, opened 2022-03-04
Missing format in the path queries

algonautshant:
Missing format in the path queries


tolikzinovyev (changes requested):
I don't think this is correct. Why msgpack? `GetAssetByID` test doesn't request msgpack.

winder (changes requested):
This is not correct, the SDKs use this flag for block and transaction exclusively. The rationale for this behavior is that those two types are on-chain while the account objects are not.

The msgpack format of account was added for TEAL/Dry-Run type operations where the messagepack format was being used for convenience as a de facto standard.

Issue #175, opened 2022-05-10
SDKs:  Support Box foreign array


michaeldiamant:
## Problem

As of https://github.com/algorand/go-algorand/issues/3890, accessing a Box requires specifying the box name in the (new) foreign Boxes array.  The story requests adding SDK support + tests for submitting Application Call transactions with a foreign Boxes array.

SDKs in-scope:
* https://github.com/algorand/go-algorand-sdk/
* https://github.com/algorand/java-algorand-sdk
* https://github.com/algorand/js-algorand-sdk
* https://github.com/algorand/py-algorand-sdk

algochoi:
SDK work for Boxes

| Work           | go-algorand | algorand-sdk-testing | go-sdk | java-sdk | py-sdk | js-sdk|
| ----------- | ------------ | --------- | --------- | -------- | --------- | --------- |
| Encode boxes as new field in application transactions  | https://github.com/algorand/go-algorand/issues/3890  | https://github.com/algorand/algorand-sdk-testing/issues/175  | 🧑‍💼  | ------ | https://github.com/algorand/py-algorand-sdk/pull/338 | ------ |
| Support box reads given app and name (algod API endpoint) | https://github.com/algorand/go-algorand/issues/3973 | ❌  | ❌ | ❌ |❌ |❌ |
| Support all box names by app (algod API Endpoint) |https://github.com/algorand/go-algorand/issues/4021 | ❌  | ❌ | ❌ |❌ |❌ |
| Support indexer endpoints for boxes | https://github.com/algorand/indexer/issues/978 | ❌  | ❌ | ❌ |❌ |❌ |

PR #181, opened 2022-05-25
Test Boxes field in application calls static array

algochoi:
This PR provides SDK tests for application transactions with the new `boxes` field:
* Added encoding tests for `boxes` in application transactions

SDK implementation:
* Python: https://github.com/algorand/py-algorand-sdk/pull/338

TODO: 
* Add integration tests for box references in application calls (Will be handled in a later PR after box read API is implemented)
* Add tests for Atomic Transaction Composer (Is this necessary?)

Tracks #175


.up-env
```diff
@@ -4,5 +4,5 @@ export TYPE="channel"
 export CHANNEL="nightly"
 
 # Used when TYPE=source
-export ALGOD_URL="https://github.com/algorand/go-algorand"
-export ALGOD_BRANCH="master"
+export ALGOD_URL="https://github.com/jannotti/go-algorand"
+export ALGOD_BRANCH="avm-box"
```
algochoi, line 8: Currently points to the fork, but should revert this back before merging to master.

features/unit/feetest.feature
```diff
@@ -5,22 +5,22 @@ Feature: Transaction fee test
...
     Then fee field is in txn
-    When I build an application transaction with operation "<operation>", application-id <application-id>, sender "<sender>", approval-program "<approval-prog-file>", clear-program "<clear-prog-file>", global-bytes <global-bytes>, global-ints <global-ints>, local-bytes <local-bytes>, local-ints <local-ints>, app-args "<app-args>", foreign-apps "<foreign-apps>", foreign-assets "<foreign-assets>", app-accounts "<app-accounts>", fee 0, first-valid <first-valid>, last-valid <last-valid>, genesis-hash "<genesis-hash>", extra-pages <extra-pages>
+    When I build an application transaction with operation "<operation>", application-id <application-id>, sender "<sender>", approval-program "<approval-prog-file>", clear-program "<clear-prog-file>", global-bytes <global-bytes>, global-ints <global-ints>, local-bytes <local-bytes>, local-ints <local-ints>, app-args "<app-args>", foreign-apps "<foreign-apps>", foreign-assets "<foreign-assets>", app-accounts "<app-accounts>", fee 0, first-valid <first-valid>, last-valid <last-valid>, genesis-hash "<genesis-hash>", extra-pages <extra-pages>, boxes "<boxes>"
     And sign the transaction
```
tzaffi, line 12: Is the plan to provide a non-trivial boxes example in this test? If not, the diff can be reduced significantly by just hard-coding an empty entry for the boxes in the steps of lines 8 and 11.

tzaffi: NVM. I see in your PR description that you are planning to add integration tests in a follow-up PR. So this must be a place-holder.

features/unit/transactions.feature
```diff
@@ -5,35 +5,37 @@ Feature: Transaction encoding
...
-      | delete    | 100            | BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4 |                                   |                                   | 0            | 0           | 0           | 0          | str:test |              |                |                                                                                                                       | 1234 | 9000        | 9010       | Mf0h6zjkEIEZPtNM3zsrg+iHQFS0fZxhgr7w35I464M= | gqNzaWfEQM26fXCfQ8Ay/uwsOya/8iJ9J9DIAfTOus2KXduNgxB195ABBCjniPNramFogzTpkmoK6s+Whj+6gpDzj1ZoFwGjdHhuiaRhcGFhkcQEdGVzdKRhcGFuBaRhcGlkZKNmZWXNBNKiZnbNIyiiZ2jEIDH9Ies45BCBGT7TTN87K4Poh0BUtH2cYYK+8N+SOOuDomx2zSMyo3NuZMQgCfvSdiwI+Gxa5r9t16epAd5mdddQ4H6MXHaYZH224f2kdHlwZaRhcHBs
```
tzaffi, line 37: These being unit tests, it seems that we ought to be able to populate the box information for these 2 new cases, no?

features/unit/feetest.feature
tzaffi: But these are unit tests, not integration, so my original question still holds. Sorry...

tzaffi:
> * Add tests for Atomic Transaction Composer (Is this necessary?)
I think this would be nice to have if not essential. Currently the step variants of "add a method call with the signing account" do not claim the ability to add reference arg arrays to the method call (i.e. accounts, foreign_apps, foreign_assets, and boxes). This seems like an important feature to test, but since the existing tests are so intricate, a smallish new scenario focusing on this aspect might make the most sense.

tzaffi (commented):
Left a couple of comments/questions.

It seems that this is a "placeholder" PR which isn't concretely asserting anything about how boxes are handled. Is this true, and if so, are we waiting on algod code in the go-algorand boxes branch in order to understand the box API, and only then can we spell out precisely what the test will be?

features/unit/feetest.feature
algochoi: Good point - I don't think the fee test would really test box features in the future so hardcoding empty entries seem better here

algochoi:
Appreciate the quick feedback:

> I think this would be nice to have if not essential. Currently the step variants of "add a method call with the signing account" do not claim the ability to add reference arg arrays to the method call (i.e. accounts, foreign_apps, foreign_assets, and boxes). This seems like an important feature to test, but since the existing tests are so intricate, a smallish new scenario focusing on this aspect might make the most sense.

I think that makes sense like an integration test for the `boxes` field.

> It seems that this is a "placeholder" PR which isn't concretely asserting anything about how boxes are handled. Is this true, and if so, are we waiting on algod code in the go-algorand boxes branch in order to understand the box API, and only then can we spell out precisely what the test will be?

Yes, the next part will be to implement the algod APIs to read from boxes. Then we can assert round-trip results (create, write, read from boxes) in the integration tests with some TEAL code.

features/unit/transactions.feature
```diff
@@ -10,32 +10,32 @@ Feature: Transaction encoding
...
-      | closeout  | 100            | BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4 |                                   |                                   | 0            | 0           | 0           | 0          | str:test | 5555,6666    | 7777,8888      | AAVDEAJ3NIYOG7XCRBKCJ3T5PUCVL2XASOP3NGX4NPPZ3UX6477PBG6E4Q,AADQIC4PMKRTFMHAAXYAFSGAUULDI2ABBIYVQJ6GZ5JHY6DJPHTU2SPHYM | 1234 | 9000        | 9010       | Mf0h6zjkEIEZPtNM3zsrg+iHQFS0fZxhgr7w35I464M= | gqNzaWfEQCgyXFm+653eOTGcswb0OPkF2NZe0nSZhSKBJzK+woWbo9dcl111qiR2xk3X0m21eTNynztPjvmloDKI10j/kAyjdHhujKRhcGFhkcQEdGVzdKRhcGFuAqRhcGFzks0eYc0iuKRhcGF0ksQgACoyATtqMON+4ohUJO59fQVV6uCTn7aa/GvfndL+5/7EIAAHBAuPYqMysOAF8ALIwKUWNGgBCjFYJ8bPUnx4aXnnpGFwZmGSzRWzzRoKpGFwaWRko2ZlZc0E0qJmds0jKKJnaMQgMf0h6zjkEIEZPtNM3zsrg+iHQFS0fZxhgr7w35I464OibHbNIzKjc25kxCAJ+9J2LAj4bFrmv23Xp6kB3mZ111Dgfoxcdphkfbbh/aR0eXBlpGFwcGw=
```
tzaffi, line 25: Ahhh - here it is! It might be a bit easier to grok these pure box changes if you put the new box data either at the beginning or the end, or create brand new test cases at the bottom. 

For an even clearer diff, consider copy-pasta'ing the original test leaving the it untouched, and just add the box fields to that test case.

tzaffi (commented):
I believe this is probably approvable but must confess to my not understanding exactly what is going on. Could you explain how the test is actually doing anything? Specifically, in `features/unit/transactions.feature` I see that one can allow `boxes` to be added to an `ApplicationCallTxn`. The hash of the txn should then reflect this new information, and this should create a totally new "golden". Is this understanding correct and are the golden's reflecting the expected transaction hash which includes boxes?

tzaffi (approved):
LGTM

features/unit/transactions.feature
```diff
@@ -5,35 +5,38 @@ Feature: Transaction encoding
...
-      | delete    | 100            | BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4 |                                   |                                   | 0            | 0           | 0           | 0          | str:test |              |                |                                                                                                                       | 1234 | 9000        | 9010       | Mf0h6zjkEIEZPtNM3zsrg+iHQFS0fZxhgr7w35I464M= | gqNzaWfEQM26fXCfQ8Ay/uwsOya/8iJ9J9DIAfTOus2KXduNgxB195ABBCjniPNramFogzTpkmoK6s+Whj+6gpDzj1ZoFwGjdHhuiaRhcGFhkcQEdGVzdKRhcGFuBaRhcGlkZKNmZWXNBNKiZnbNIyiiZ2jEIDH9Ies45BCBGT7TTN87K4Poh0BUtH2cYYK+8N+SOOuDomx2zSMyo3NuZMQgCfvSdiwI+Gxa5r9t16epAd5mdddQ4H6MXHaYZH224f2kdHlwZaRhcHBs
```
jasonpaulos, line 27: These tests look good, but it would be great to see a test for an empty box name as well

jasonpaulos (approved):
The encoding tests look good, I just left a comment about another case I'd like to see. I'll approve since this is going into a feature branch, so not everything needs to be complete yet

features/unit/transactions.feature
algochoi: Good point - added a test case for appID=0 and empty box name below this test case

jasonpaulos: Thanks!

PR #185, opened 2022-06-07
adding get method tests

barnjamin:


features/unit/abijson.feature
```diff
@@ -86,3 +86,26 @@ Feature: AbiJson
...
+    Examples:
+      | methodsig                | name | error |
+      | add(uint32,uint32)uint32 | add  |       |
+      | add(uint32,uint32)uint32 | sub  | found 0 methods |
+
+  Scenario Outline: Get a method by name from Contract
+    When I create the Method object from method signature "<methodsig>"
+    And I create a Contract object from the Method object
+    And I get the method from the Contract by name "<name>"
+    Then the produced method signature should equal "<methodsig>" if there is an error it is equal to "<error>"
```
algochoi, line 106: I'd just change this to:
```suggestion
    Then the produced method signature should equal "<methodsig>". If there is an error it is "<error>"
```

Just so it is consistent with the other error check steps (and easily grep-able)

algochoi (commented):
Looking good - just wondering what would happen if the method name is not unique and if there is a test case for that.

features/unit/abijson.feature
```diff
@@ -86,3 +86,26 @@ Feature: AbiJson
...
+  Scenario Outline: Get a method by name from Interface
+    When I create the Method object from method signature "<methodsig>"
+    And I create an Interface object from the Method object
+    And I get the method from the Interface by name "<name>"
+    Then the produced method signature should equal "<methodsig>" if there is an error it is equal to "<error>"
+
+    Examples:
+      | methodsig                | name | error |
+      | add(uint32,uint32)uint32 | add  |       |
+      | add(uint32,uint32)uint32 | sub  | found 0 methods |
```
ahangsu, line 100: nit: is it possible to test adding dupe name methods in interface/contract, such that we can see the error raised? feel free to ignore if that is getting complex.

ahangsu (commented):
just one comment

algochoi (approved):
This looks good to me :shipit:

features/unit/abijson.feature
```diff
@@ -86,3 +86,30 @@ Feature: AbiJson
...
+
+    Examples:
+      | methodsig                | extramethod              | name | error           |
+      | add(uint32,uint32)uint32 |                          | add  |                 |
+      | add(uint32,uint32)uint32 |                          | sub  | found 0 methods |
+      | add(uint32,uint32)uint32 | add(uint64,uint64)uint64 | add  | found 2 methods |
+
+  Scenario Outline: Get a method by name from Contract
+    When I create the Method object from method signature "<methodsig>"
+    And I create another Method object from method signature "<extramethod>"
```
tzaffi, line 106:
```suggestion
    When I append to my Method objects list in the case of a non-empty signature "<methodsig>"
    And I append to my Method objects list in the case of a non-empty signature "<methodsig>"
```

nit: we can consolidate the add/create method step definitions into a single one. See also my upcoming comment on the adjoining py-sdk PR.

features/unit/abijson.feature
```diff
@@ -86,3 +86,30 @@ Feature: AbiJson
...
+    And I create a Contract object from the Method object
+    And I get the method from the Contract by name "<name>"
+    Then the produced method signature should equal "<methodsig>". If there is an error it begins with "<error>"
```
tzaffi, line 109: This step suggests that our ABI Contract code has a bug in it. How can you even get to this point? Shouldn't we blow up as soon as we attempt to construct a contract with a duplicate error name or add such a duplicate?

@algochoi - any thoughts?

barnjamin: dupe name is fine since it generates a diff signature (assuming args are different)

features/unit/abijson.feature
barnjamin: I need it to be a distinct column so I can test same name but different selector

features/unit/abijson.feature
tzaffi: Thinking more... overloading seems to be ok according to ARC-4. The only issue would be if the signatures have the same selector (which in practice means they have the same signature). So I guess this is ok, but then we should allow querying a method by its signature, no?

features/unit/abijson.feature
tzaffi: Right, I meant to make the second one take `<extramethod>` ... fixed

features/unit/abijson.feature
barnjamin: we _could_ but I don't expect it will be common. And someone can implement their own method getter thing since the method list is public in all interface/contracts.

algochoi: > So I guess this is ok, but then we should allow querying a method by its signature, no?

I think this also makes sense. But if we don't see it being a common use case, the current approach seems fine.

features/unit/abijson.feature
```diff
@@ -89,9 +89,9 @@ Feature: AbiJson
 
 
   Scenario Outline: Get a method by name from Interface
-    When I create the Method object from method signature "<methodsig>"
-    And I create another Method object from method signature "<extramethod>"
-    And I create an Interface object from the Method object
+    When I append to my Method objects list in the case of a non-empty signature "<methodsig>"
```
tzaffi, line 92: Thinking more about it, you might be able to reuse the old instruction `I create the Method object from method signature...` if it also appends to the list... but for clarity and easier backwards compatibility, this new step is probably preferred.

tzaffi (approved):
LGTM but we can't really merge this until all the SDK's implement. If you want to merge earlier, you can add a special tag to the scenarios such as `@abi.contract` which will cause the scenario to be skipped in non-implementing SDK's.

jasonpaulos (approved):
Looks good, thanks for making tests

features/unit/abijson.feature
```diff
@@ -86,3 +86,30 @@ Feature: AbiJson
...
+    And I append to my Method objects list in the case of a non-empty signature "<extramethod>"
+    And I create a Contract object from my Method objects list
+    And I get the method from the Contract by name "<name>"
+    Then the produced method signature should equal "<methodsig>". If there is an error it begins with "<error>"
+
+    Examples:
+      | methodsig                | extramethod              | name | error           |
+      | add(uint32,uint32)uint32 |                          | add  |                 |
+      | add(uint32,uint32)uint32 |                          | sub  | found 0 methods |
+      | add(uint32,uint32)uint32 | add(uint64,uint64)uint64 | add  | found 2 methods |
```
tzaffi, line 115:
```suggestion
      | add(uint32,uint32)uint32 | add(uint64,uint64)uint64 | add  | found 2 methods |

```

nit: space

PR #189, opened 2022-06-14
Assert that Atomic Transaction Composer's method adder checks for arg-length consistency

tzaffi:
# Method Adder arg-length Check Assertion

Adding a new integration test _Scenario Outline_ that asserts that when adding an ABI method via the Atomic Transaction Composer, a validity check is undertaken to ensure that the number of runtime args is the number that is expected by the method's signature. In particular, a new step is introduced, which is a variant of the "I add a method call with..." step family.

>  I add a method call with the signing account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments; any resulting exception has key "{none-or-exception-key}".

Will not merge until all 4 SDK's implement the new step, or will at least add a new tag before merging.

## SDK Specific Situations

Please see #190 for some cursory investigations that were conducted on how each SDK's `AtomicTransactionComposer` asserts that the number of arguments provided to the app call, match the number according the method signature.

### DIFF
--- a/features/integration/abi.feature
+++ b/features/integration/abi.feature
@@ -18,15 +18,16 @@ Feature: ABI Interaction
     When I create the Method object from method signature "<method-signature>"
     And I create a new method arguments array.
     And I append the encoded arguments "<app-args>" to the method arguments array.
-    And I add a method call with the transient account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments.
     # When "none" is provided for <none-or-exception-pattern> there should be no exception, otherwise, the error's message should satisfy the regex
-    Then the most recently added method call has an exception which satisfies "<none-or-exception-pattern>".
+    And I add a method call with the transient account, the current application, suggested params, on complete "noop", current transaction signer, current method arguments; any resulting exception satisfies the regex "<none-or-exception-pattern>".
 
     Examples:
       | method-signature         | app-args                               | none-or-exception-pattern |
       | add(uint64,uint64)uint64 | AAAAAAAAAAE=,AAAAAAAAAAI=              | none                      |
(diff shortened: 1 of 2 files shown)


features/integration/abi.feature
```diff
@@ -12,6 +12,34 @@ Feature: ABI Interaction
     And I wait for the transaction to be confirmed.
     And I remember the new application ID.
 
+  @z
```
tzaffi, line 15:
```suggestion
```

features/integration/abi.feature
```diff
@@ -12,6 +12,33 @@ Feature: ABI Interaction
     And I wait for the transaction to be confirmed.
     And I remember the new application ID.
 
+  Scenario Outline: Method call exceptions
```
jasonpaulos, line 15: Also, since this scenario only tests for the incorrect number of arguments, I'd recommend renaming it to be more specific. The reason being this doesn't test every type of error that could happen, e.g. passing the wrong type is something not tested, and I think that would be difficult to test with the current steps.

jasonpaulos (approved):

tzaffi:
Thanks for approving @jasonpaulos. I cannot merge _as is_ since it would fail the SDK tests. @michaeldiamant what do you think. Should I add a new tag `@unit.atcMethodArgs` (or something like that) and then merge? Or should I leave it open till all the SDK's have implemented?

michaeldiamant:
> Should I add a new tag @unit.atcMethodArgs (or something like that) and then merge? Or should I leave it open till all the SDK's have implemented?

@tzaffi I'd favor a new tag.  Proliferating tags is a concern, but longstanding. open PRs feels to me like a worse tradeoff.

PR #191, opened 2022-06-15
Remove empty box name from transaction encoding tests

algochoi:
This PR removes the empty box name test case from the encoding tests (Boxes no longer support empty names - see [this PR](https://github.com/algorand/go-algorand/commit/df49acead59403edbe136e4b2c51635548643a72)).


jannotti:
Transactions _should_ be able to contain empty box names, even though that she's not convey any ability to touch such a cuz in the curse if the transaction. That's how a transaction can indicate that it wants more IO bandwidth for boxes.

algochoi:
Closing this PR - thanks for the clarification around empty box name encoding

PR #193, opened 2022-06-18
indexer-live: allow override of git info

tzaffi:
The only currently proposed change is to allow controlling the repo and branch of the `sdk-harness-indexer-live` docker image. See the [companion Python SDK PR](https://github.com/algorand/py-algorand-sdk/pull/352/files) for greater context.

Some questions also arose regarding the staleness of other indexer docker images.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -95,7 +95,6 @@ services:
       dockerfile: ./docker/indexer/Dockerfile
       args:
         URL: "https://github.com/algorand/indexer"
-        # Can we follow up on this now?
         # TODO: Set back to master when include-all makes it to master.
         BRANCH: "develop"
         SHA: "cf93e3acacdf6fde9afd0b6b24fa0fde723ff43b"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,8 +30,9 @@ services:
       context: .
       dockerfile: ./docker/indexer/Dockerfile
       args:
-        URL: "https://github.com/algorand/indexer"
-        BRANCH: "develop"
+        # allow override of the git information.
+        URL: "${INDEXER_URL:-https://github.com/algorand/indexer}"
+        BRANCH: "${INDEXER_BRANCH:-develop}"
     ports:
       - 60002:8980
     restart: unless-stopped
@@ -53,6 +54,7 @@ services:
       args:
         URL: "https://github.com/algorand/indexer"
         BRANCH: "master"
+        # Is there a reason why we're still pinning to a particular commit?
         SHA: "a30878e3669310c30a2b916fb41511516b906c9a"
     ports:
       - 59999:8980
@@ -74,6 +76,7 @@ services:
       args:
         URL: "https://github.com/algorand/indexer"
         BRANCH: "master"
+        # Is there a reason why we're still pinning to a particular commit?
         SHA: "a30878e3669310c30a2b916fb41511516b906c9a"
     ports:
       - 59998:8980
@@ -94,6 +97,7 @@ services:
       dockerfile: ./docker/indexer/Dockerfile
       args:
         URL: "https://github.com/algorand/indexer"
+        # Can we follow up on this now?
         # TODO: Set back to master when include-all makes it to master.
         BRANCH: "develop"
         SHA: "cf93e3acacdf6fde9afd0b6b24fa0fde723ff43b"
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -53,6 +54,7 @@ services:
       args:
         URL: "https://github.com/algorand/indexer"
         BRANCH: "master"
+        # Is there a reason why we're still pinning to a particular commit?
```
tzaffi, line 57: @winder  - I think you understand the context pretty well here. Do you think we should still be pinning to specific commits? Could you have a look at the other comments as well?

winder: It's been a long time since I looked at this. Do you have a recommendation for how this could be improved?

If I remember correctly, versions are pinned strategically to prevent new tests/features from breaking old releases. For example, if an endpoint were deprecated or a new field was added. i.e. `indexer-221-2` is something like version 2.2.1.

If we are going to improve this framework, I think we should push to get production docker containers.

@algojohnlee are we going to have production docker containers soon? [It would be great if we could use this as a solid foundation.](https://hub.docker.com/r/wwinder/algod)

docker-compose.yml
```diff
@@ -74,6 +76,7 @@ services:
...
```
winder, line 79: Yes, this is pinned to a specific release. Based on the comment I guess this is the one that introduced Applications.

docker-compose.yml
```diff
@@ -94,6 +97,7 @@ services:
       dockerfile: ./docker/indexer/Dockerfile
       args:
         URL: "https://github.com/algorand/indexer"
+        # Can we follow up on this now?
```
winder, line 100: Yes, but it would still be pinned to the commit mapping to the 2.3.x release. IIRC this is the release where we started importing go-algorand.

winder:
Interestingly, when [trying to list all the major tech debt I could think of](https://algorand.atlassian.net/wiki/spaces/LAMPREY/pages/2364375047/Technical+Debt), this is the only one I had for the SDK testing:

* SDK Testing Framework is hardcoded to use the nightly release channel. It has custom docker containers. It is hard to run locally.
Impact:
  *    It is difficult to use and maintain.
  *   There are sometimes false-negative failures due to the framework which waste developer time.

There main remaining problem I see is that the developer still needs to figure out the right versions for Indexer and Algod. Maybe thats a necessary evil since you would, presumably, have branches in both repos supporting a new feature.

docker-compose.yml
```diff
@@ -30,8 +30,9 @@ services:
       context: .
       dockerfile: ./docker/indexer/Dockerfile
       args:
-        URL: "https://github.com/algorand/indexer"
-        BRANCH: "develop"
+        # allow override of the git information.
+        URL: "${INDEXER_URL:-https://github.com/algorand/indexer}"
+        BRANCH: "${INDEXER_BRANCH:-develop}"
```
tzaffi, line 35: this is the only actual change

docker-compose.yml
tzaffi, line 57:
```suggestion
```

docker-compose.yml
tzaffi, line 79:
```suggestion
```

docker-compose.yml
tzaffi, line 100:
```suggestion
```

docker-compose.yml
```diff
@@ -116,6 +120,7 @@ services:
       dockerfile: ./docker/indexer/Dockerfile
       args:
         URL: "https://github.com/algorand/indexer"
+        # Can we follow up on this now?
```
tzaffi, line 123:
```suggestion
```

docker-compose.yml
michaeldiamant: @tzaffi @winder Reading the thread it's unclear to me if the `indexer-221-X` images can be removed.  What's the verdict?

Provided it's _safe_, removing these images provides an immediate build time reduction.  Obviously, published containers are a welcomed future improvement.

winder:
GitHub trick: On the files tab, you can batch suggestions together and add them all in a single commit.

tzaffi:
👍

docker-compose.yml
tzaffi: I'm happy to open another PR that aims at understanding which tests are impacted and -with support- removes such tests. Ok?

jdtzmn (approved):

michaeldiamant (approved):

PR #199, opened 2022-06-24
SourceMap decoder tests

barnjamin:


features/unit/sourcemap.feature
```diff
@@ -0,0 +1,22 @@
+@unit
+Feature: SourceMap
+
+    @unit.sourcemap.pctoline
+    Scenario Outline: Parse a SourceMap from json and get the line corresponding to a PC
+        Given a source map json file "<sourcemap-file>"
+        Then getting the line that corresponds to the PC "<pc-value>" produces "<line-number>"
+
+        Examples:
+            | sourcemap-file | pc-value | line-number |
```
algochoi, line 10: Thanks for making the tests! 

One thought I had was since it's easy to loop and append strings in the SDKs anyway, we could just check against an entire string like: `0:0,1:2,2:2,3:3,4:3,5:3,6:4,7:5,8:6,9:8,10:8,11:9,12:9,13:9,14:10,15:11,16:12`

In the SDK, it could generate the string and compare against it:
```
decoded_map = source_map.SourceMap(sm)
result_string = ''
for k, v in decoded_map.pc_to_line.items():
    result_string += f'{k}:{v},'
print(result_string[:-1])
```

barnjamin: good point, I'll implement this

features/resources/v2algodclient_responsejsons/sourcemap.json
```diff
@@ -0,0 +1 @@
+{"version":3,"sources":["quine.teal"],"names":[],"mapping":";AAOA;;;;;;;;;;;;;;;;;;;;;;;;;;;;AACA;AACA;;;AACA;;;AACA;AACA;;AACA;AACA;;;AACA;AACA;AACA;;AACA;;AACA;AACA;AACA","mappings":";AAOA;;;;;;;;;;;;;;;;;;;;;;;;;;;;AACA;AACA;;;AACA;;;AACA;AACA;;AACA;AACA;;;AACA;AACA;AACA;;AACA;;AACA;AACA;AACA"}
```
michaeldiamant, line 1: @barnjamin Asking without having research under the assumption that you know off-hand:  What requires duplicating `sourcemap.json` in 2 locations rather than retaining a single copy?

barnjamin: I don't know in fact, mostly doing it out of superstition

michaeldiamant: Gotcha - I imagine only 1 copy is needed.  I'd like to leave the thread open as a visual reminder to investigate.

michaeldiamant: @barnjamin Based on a quick review of Python + Go SDKs, the resource loading method references `features/resources`.

The finding suggests to me that `features/unit/v2algodclient_responsejsons/sourcemap.json` can be removed.

If you agree with the rationale, can you remove `features/unit/v2algodclient_responsejsons/sourcemap.json` and confirm downstream builds pass?

features/unit/sourcemap.feature
```diff
@@ -0,0 +1,11 @@
+@unit
+Feature: SourceMap
+
+    @unit.sourcemap.pctoline
+    Scenario Outline: Parse a SourceMap from json and compare with expected pc:line string
+        Given a source map json file "<sourcemap-file>"
+        Then the string composed of pc:line number equals "<pc-to-line>"
```
algochoi, line 7: Looks good, just for extra peace of mind, I'd test the line:pc mapping as well, since it is exposed to the public.

features/integration/compile.feature
```diff
@@ -20,3 +21,11 @@ Feature: Compile
       | "programs/one.teal"             | "programs/one.teal.tok"             |
       | "programs/zero.teal"            | "programs/zero.teal.tok"            |
       | "programs/abi_method_call.teal" | "programs/abi_method_call.teal.tok" |
+
+  @compile.sourcemap
+  Scenario Outline: Teals return a valid Source Map
```
algochoi, line 26: nit: I wonder why the previous tests here chose to write `Teals`

michaeldiamant: Agree - seems a bit strange to me.  At least for the highlighted scenario, I imagined something like "Algod compiling TEAL returns a valid source map" or similar.  

I don't feel strongly.

algochoi (approved):
LGTM, thanks for the changes!

features/unit/sourcemap.feature
```diff
@@ -0,0 +1,35 @@
+@unit.sourcemap
+Feature: SourceMap
+
+    @unit.sourcemap.full
```
michaeldiamant, line 4: @barnjamin Are the following subtags needed?  
* `@unit.sourcemap.full`
* `@unit.sourcemap.pctoline`
* `@unit.sourcemap.linetopc`

The downstream SDKs rely on `@unit.sourcemap`.  Rationale for removing:  Presents fewer chances for downstream SDKs to be out-of-sync (e.g. miss running a test).

It's possible I'm missing the rationale for retaining them.  Feel welcomed to suggest retaining the tags.

barnjamin: You're right, I've removed all but the top tag

PR #207, opened 2022-07-06
Updating url to newer updater

algojack:


algobarb (approved):
nice. Maybe we need a checklist of places we need to update if we make a significant updater change.

barnjamin (approved):
Will this need to change with any frequency going forward?

PR #210, opened 2022-07-11
Boxes: Point go-algorand to boxes_response_rework

michaeldiamant:
Extends #208 to point go-algorand branch to https://github.com/algorand/go-algorand/pull/4249.  Enables SDKs to make corresponding API changes.

### DIFF
--- a/.up-env
+++ b/.up-env
@@ -5,4 +5,4 @@ export CHANNEL="nightly"
 
 # Used when TYPE=source
 export ALGOD_URL="https://github.com/algorand/go-algorand"
-export ALGOD_BRANCH="feature/avm-box"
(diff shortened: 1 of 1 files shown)


algochoi (commented):
How about we just wait for the `boxes_response_rework` changes to be merged into the current feature branch? I'm wondering if there is urgency to create a new PR (instead of local branch testing), since we'd have to revert it back when the parent PR needs to be merged in.

michaeldiamant:
> How about we just wait for the boxes_response_rework changes to be merged into the current feature branch? 

@algochoi Sounds good - I'm on-board to avoid merging the PR.  Minimally, I opened the PR for visibility purposes.

PR #216, opened 2022-08-02
Add unit tests for state proof algod endpoints.

winder:
Tests for the new state proof endpoints. These were captured by starting a private network.

Responses from Indexer with the updated transaction and block header responses are still required.


features/resources/stateproof/v1_GetBlock_txn.json
```diff
@@ -0,0 +1,272 @@
...
+    "txn256": "ME1NR8Z+cqGCYrgcbW8vjFfzt7TKaOraexjP6SiQw0g=",
+    "txns": [
+      {
+        "txn": {
+          "fv": 821,
+          "lv": 1821,
+          "snd": "XM6FEYVJ2XDU2IBH4OT6VZGW75YM63CM4TC6AV6BD3JZXFJUIICYTVB5EU",
+          "sp": {
+            "P": {
+              "hsh": {
```
algoidan, line 32: I wonder why the proofs are empty.
I would expect to have at least one

features/resources/stateproof/v1_GetProof.json
```diff
@@ -0,0 +1,7 @@
+{
+  "hashtype": "sha512_256",
+  "idx": 0,
+  "proof": "",
+  "stibhash": "1mr5NAjTaheChqu/6dt5JR8yHC8pZW4w8fgOcfSoHEU=",
+  "treedepth": 0
```
algoidan, line 6: should we take a non-empty block example?

algoidan (commented):
overall looks pretty good.
I  would suggest maybe to more complex examples if possible

features/resources/stateproof/v1_GetBlock_txn.json
winder: These are generated when I start a network with sandbox, do you have a more complex example?

 The SDKs use msgpack for this endpoint so I need to convert this from json to msgp also.

features/resources/stateproof/v1_GetProof.json
winder: There was a transaction in the block I queried, what else should be in this response?

Eric-Warehime (approved):

features/resources/stateproof/v1_GetBlock_txn.json
algoidan: Taking a second look here, I'm pretty sure that this represents invalid proof. The depth is 1 and there is no digest in the slice. However, Proofs related to the merkle signature scheme look fine.
I suggest we look at it together

PR #220, opened 2022-08-03
Enhancement: Remove the dataset indexer images and the indexer.feature + utilize sandbox for algod/indexer/postgres containers

tzaffi:
# Summary
After removing the 4 non-live indexer instances and realizing that the live one wasn't operational, it was easier to swap out all the docker containers with the much more stable sandbox.

## TODO

- [x] Get consensus on the best way to move forward with this PR. In particular, I see a few approaches:
  - [x] **Most Aggressive** - ~Stop testing all v1 endpoints. This will require _significant_ investigation into what the v1 endpoint-dependencies are inside of SDK steps implementations and implicitly their features.~
  - [x] **Medium** - Continue with the current PR and leave the _V1 Endpoint Problem_ to a future PR. The current approach requires the step removal/modifications described below.
  - [x] **Least Disruptive** - ~instead of removing features and steps -as is done in this PR- re-tag everything so that obsolete steps are marked `@obsolete`. This requires much more detailed tagging of steps.~
  - [x] ~Something else?~
- [ ] Have all SDK's removed their step dependency on indexer v1? If so, this PR can be merged. Details:
  - [ ] [`@rekey_v1` and `@send` tags] - `then("the transaction should go through")` **DO NOT `assert "type" in context.acl.transaction_by_id(context.txn.get_txid())` because this relies on indexer v1**
  - [ ] [`@applications.verified` tag] - `step("I wait for the transaction to be confirmed.")` **DO NOT `assert "type" in context.acl.transaction_by_id(context.app_txid)`**

## Related PR's
* [Sandbox](https://github.com/algorand/sandbox/pull/136)
* Example usage in [Python SDK](https://github.com/algorand/py-algorand-sdk/pull/367)
  * This includes a [high-level explanation](https://github.com/algorand/py-algorand-sdk/pull/367#issue-1327887064) of `up.sh`'s logic (see step `A.iv.`)

## Files introduced/modified/dropped:
* Introducing `.env` to help configure the sandbox. This replaces `.up-env`
* `config.harness` is a sandbox template used for standing it up
* `docker-compose.yml` and `docker/***`are removed as they are subsumed by sandbox's docker build artifacts
* `features/integration/` - removed integration test features: 
  * `evaldelta.feature` - this hasn't been running since February 2022
  * `indexer.feature` and `features/resources/v23x*/***` fixtures - this was a static test last modified March 2021 and which didn't interact with algod
* `network_config/` - removing all configurations for bootstrapping the algod node. As Sandbox is now in charge of bringing up the network, all configurations are "owned" by it. If in the future, a network that doesn't exist in Sandbox is required by SDK testing, it is recommended that this network be introduced to Sandbox via a pull request.
* `scripts/` - restructured the test scripts:
  * `restart.sh` - this is removed as sandbox already has `./sandbox up` which serves the same purpose in a robust fashion
  * `down.sh` - delegates to `./sandbox down` and `./sandbox clean`
  * `up.sh` - serves a similar purpose as before, delegating logic based on `channel` / `source` configuration and also populating the `config.harness` template to be utilized in `/sandbox up config.harness`
* `utils/` - removing all files except for `utils/implements_all_tests.sh`. The removed files include:
  * `create_indexer_integration_goldens.sh` test case generator that depended on the non-live indexers being removed
  * `utils/create_transactions.sh` an old test case generating script. 
  * `*teal` and `*tok` files which weren't used in any tests. These have been subsumed by the programs in `features/resources/programs/`
* `utils/unused_steps_analysis.ipynb` - Script which uses a gherkin parser to aggregate all feature steps, and then uses regular expressions and fuzzy string logic to help discover unused steps in each SDK. This helped discover an average of 12 obsolete steps in each of the SDK's. If we were to start running nightly jobs in this repo, and with a bit more TLC, this could be automated to alert of unused steps.

## Steps Removal Guide
The following steps have been orphaned due to the removal of `indexer.feature` and `evaldelta.feature`:

1. `when('I use {indexer} to lookup account "{account}" at round {round}')`
2. `when('I use {indexer} to search for an account with {assetid}, {limit}, {currencygt}, {currencylt}, "{auth_addr:MaybeString}", {application_id}, "{include_all:MaybeBool}" and token "{token:MaybeString}"')`
3. `when('I use {indexer} to search for an account with {assetid}, {limit}, {currencygt}, {currencylt}, "{auth_addr:MaybeString}", {application_id} and token "{token:MaybeString}"')`
4. `when('I use {indexer} to search for an account with {assetid}, {limit}, {currencygt}, {currencylt} and token "{token:MaybeString}"')`
5. `when("I get the next page using {indexer} to search for an account with {assetid}, {limit}, {currencygt} and {currencylt}")`
6. `when('I use {indexer} to search for applications with {limit}, {application_id}, "{include_all:MaybeBool}" and token "{token:MaybeString}"')`
7. `when('I use {indexer} to search for applications with {limit}, {application_id}, and token "{token:MaybeString}"')`
8. `when('I use {indexer} to lookup application with {application_id} and "{include_all:MaybeBool}"')`
9. `when("I use {indexer} to lookup application with {application_id}")`
10. `when("I get the next page using {indexer} to lookup asset balances for {assetid} with {currencygt}, {currencylt}, {limit}")`
11. `when("I use {indexer} to search for all {assetid} asset transactions")`
12. `when('I use {indexer} to search for all "{accountid}" transactions')`
13. `when("I use {indexer} to check the services health")`
14. `when("I use {indexer} to lookup block {number}")`
15. `when('I use {indexer} to lookup asset balances for {assetid} with {currencygt}, {currencylt}, {limit} and token "{token}"')`
16. `when("I use {indexer} to lookup asset {assetid}")`
17. `when("I get the next page using {indexer} to search for transactions with {limit} and {maxround}")`
18. `when('I use {indexer} to search for transactions with {limit}, "{noteprefix:MaybeString}", "{txtype:MaybeString}", "{sigtype:MaybeString}", "{txid:MaybeString}", {block}, {minround}, {maxround}, {assetid}, "{beforetime:MaybeString}", "{aftertime:MaybeString}", {currencygt}, {currencylt}, "{address:MaybeString}", "{addressrole:MaybeString}", "{excludecloseto:MaybeString}" and token "{token:MaybeString}"')`
19. `when('I use {indexer} to search for transactions with {limit}, "{noteprefix:MaybeString}", "{txtype:MaybeString}", "{sigtype:MaybeString}", "{txid:MaybeString}", {block}, {minround}, {maxround}, {assetid}, "{beforetime:MaybeString}", "{aftertime:MaybeString}", {currencygt}, {currencylt}, "{address:MaybeString}", "{addressrole:MaybeString}", "{excludecloseto:MaybeString}", {application_id} and token "{token:MaybeString}"')`
20. `when('I use {indexer} to search for assets with {limit}, {assetidin}, "{creator:MaybeString}", "{name:MaybeString}", "{unit:MaybeString}", and token "{token:MaybeString}"')`
21. `when('indexer client {index} at "{address}" port {port} with token "{token}"')`
22. `then("I get transactions by address only")` - needs indexer v1 [`@algod` tag]
23. `then("I can get the transaction by ID")` - needs indexer v1 [`@algod` tag]
24. `when("I get recent transactions, limited by {cnt} transactions")` - [`@algod` tag]
25. `then("I get transactions by address and date")` - [`@algod` tag]

### Additional `indexer.feature` steps which were previously overlooked, but need removal.
These are uniquely discoverable (unless noted otherwise) by the following search strings:
26. `I receive status code`
27. `transactions, has the previous block hash`
28. `has a frozen status of`
29. `with a total amount of`
30. `μalgos`
31. `The asset found has:`
32. `with the asset, the first`
33. `, the first has`
34. `The first account is online and has`
35. `transactions in the response, the first is`
36. `Every transaction has tx-type`
37. `Every transaction has sig-type`
38. `Every transaction has round >=`
39. `Every transaction has round <=`
40. `Every transaction has round`
41. `Every transaction works with asset-id`
42. `Every transaction is older than`
43. `Every transaction is newer than`
44. `Every transaction moves between`
45. `assets in the response, the first is`
46. `the parsed response should equal` - **CAREFUL!!! there are unit as well as integration test versions. Make sure to only remove the indexer integration test version**
47. `the unconfirmed pending transaction by ID` (actually an `evaldelta` feature)
48. `the confirmed pending transaction by ID`

## Steps being modified (for example in the Python SDK)
* [`@rekey_v1` and `@send` tags] - `then("the transaction should go through")`
  * DO NOT `assert "type" in context.acl.transaction_by_id(context.txn.get_txid())` because this relies on indexer v1
* [`@applications.verified` tag] - `step("I wait for the transaction to be confirmed.")`
  * DO NOT `assert "type" in context.acl.transaction_by_id(context.app_txid)`


docker-compose.yml
```diff
@@ -24,129 +24,7 @@ services:
       - sdk-harness
     volumes:
       - genesis-file:/genesis-file
```
tzaffi, line 26: Removing all indexer dockers for now. We should bring them back in when we have a live test that uses them.

features/integration/evaldelta.feature
```diff
@@ -1,26 +0,0 @@
-Feature: EvalDelta
```
tzaffi, line 1: This feature was only running on the Java SDK, and even on Java, it [hasn't been running since Feb 9, 2022](https://github.com/algorand/java-algorand-sdk/pull/290/files).

scripts/up.sh
```diff
@@ -63,17 +23,58 @@ else
...
+
+rootdir=$(dirname "$0")
+pushd "$rootdir"/.. > /dev/null || exit
+
+git clone --branch "$SANDBOX_BRANCH" --single-branch "$SANDBOX_URL" "$SANDBOX_DIR"
+
+echo "up.sh. seconds it took to get to end of cloning sandbox into $SANDBOX_DIR: " + $(($(date "+%s") - $START))
+
+echo "Bringing up network with '$TYPE' configuration."
+cp .env "$SANDBOX_DIR"/.
```
tzaffi, line 75: TODO: do I still need this `.env`? Seems like I'm duplicating configuration and that ought to be able to just stick with `config.harness` generation and remove the dummy `.env` just added to sandbox.

config.harness
```diff
@@ -0,0 +1,16 @@
+export ALGOD_CHANNEL="#ALGOD_CHANNEL"
```
tzaffi, line 1: Template used for generating config used by the `./sandbox up $CONFIG` command.

README.md
```diff
@@ -49,17 +50,13 @@ These reside in the [integration features directory](features/integration)
 | ---------------------- | -------------------------------------------------------------------------------------- |
 | @abi                   | Test the Application Binary Interface (ABI) with atomic txn composition and execution. |
 | @algod                 | General tests against algod REST endpoints.                                            |
-| @application.evaldelta | Test that eval delta fields are included in algod and indexer.                         |
```
tzaffi, line 52: Propose the removal of the `evaldelta` feature as it hasn't been running anywhere since February 2022.

README.md
```diff
@@ -49,17 +50,13 @@ These reside in the [integration features directory](features/integration)
...
 | @applications.verified | Submit all types of application transactions and verify account state.                 |
 | @assets                | Submit all types of asset transactions.                                                |
 | @auction               | Encode and decode bids for an auction.                                                 |
 | @c2c                   | Test Contract to Contract invocations and injestion.                                   |
 | @compile               | Test the algod compile endpoint.                                                       |
 | @dryrun                | Test the algod dryrun endpoint.                                                        |
 | @dryrun.testing        | Test the testing harness that relies on dryrun endpoint. Python only.                  |
-| @indexer               | Test all types of indexer queries and parameters against a static dataset.             |
-| @indexer.231           | REST Client Integration Tests for Indexer 2.3.1+                                       |
-| @indexer.applications  | Endpoints and parameters added to support applications.                                |
```
tzaffi, line 62: Propose the removal of all indexer features which only rely on hard-coded database fixtures.

README.md
```diff
@@ -72,9 +69,9 @@ However, a few are not fully supported:
 
 | tag                             | SDK's which implement        |
 | ------------------------------- | ---------------------------- |
-| @application.evaldelta          | Java only                    |
 | @dryrun.testing                 | Python only                  |
-| @indexer.rekey                  | missing from Python and JS   |
```
tzaffi, line 77: TODO: ^^^ does this actually exist ? ^^^

scripts/up.sh
```diff
@@ -63,17 +23,63 @@ else
...
 
-echo "Bringing up network with '$TYPE' configuration."
+echo "Before bootrapping, try cleaning up first..."
+
 
-# Make sure it isn't running
-./scripts/down.sh
+pwd
+ls -a
+echo "^^^^CWD STATUS FROM up.sh^^^^"
```
tzaffi, line 31: TODO: clean up the debug printouts in this script

scripts/up.sh
tzaffi: I got convinced that I need it, mainly because [this conversation](https://github.com/algorand/py-algorand-sdk/pull/367#discussion_r941331468).

TLDR - this repo should be able to standup a test-harness with a simple call to `up.sh`. This might even become a github-actions based integrations test.

.env
```diff
@@ -0,0 +1,31 @@
...
+# Used when TYPE==channel
+ALGOD_CHANNEL="nightly"
+
+# Used when TYPE==source
+ALGOD_URL="https://github.com/algorand/go-algorand"
+ALGOD_BRANCH="master"
+ALGOD_SHA=""
+
+# Used regardless of TYPE
+NETWORK_TEMPLATE="images/algod/DevModeNetwork.json"
```
algochoi, line 12: Since we use the sandbox network templates, it looks like we could delete the `network_config` directory in this repo

tzaffi: This [build](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/805/workflows/302732f3-ee1e-4fff-866d-4c814be3abdd/jobs/3377) passes on py-sdk after the deletion you suggested, so am resolving this. Thanks!!!!

tzaffi: FYI - I've added this explanatory comment:

image

README.md
tzaffi: @winder @jasonpaulos ^^^

README.md
tzaffi: @winder ^^^

README.md
```diff
@@ -72,9 +69,9 @@ However, a few are not fully supported:
...
+| @unit.c2c                       | Not python... which one???   |
```
tzaffi, line 73: ^^^ TODO: figure this out ^^^

.env
```diff
@@ -0,0 +1,31 @@
+TYPE="source" # OR "channel"
```
tzaffi, line 1: This replaces the former `.up_env`, is used both in `up.sh` and `down.sh` and overwrites the new sandbox `.env` during `up.sh`

scripts/up.sh
```diff
@@ -1,79 +1,79 @@
...
 
 set -e
 
-rootdir=`dirname $0`
-pushd $rootdir/.. > /dev/null
+START=$(date "+%s")
 
 # Defaults
 TYPE_OVERRIDE=""
-DAEMON_FLAG="-d"
```
tzaffi, line 12: Removing most of the interactively configurable functionality of `up.sh` (and `down.sh` as well). I suspect -with no evidence- that this was rarely used. Most of the safe functionality can be achieved by setting variables in `.env`. If there is a specific capability that is lost due to this change and should be kept, please provide feedback.  We can discuss how to keep the functionality- whether by exposing int `.env` or bringing back more interactivity in this script.

@winder @michaeldiamant

scripts/up.sh
```diff
@@ -1,79 +1,79 @@
...
-if [[ $SKIP_BUILD -eq 0 ]]; then
-  docker-compose build --no-cache $PARALLEL
+if [[ $OSTYPE == 'darwin'* ]]; then
+  SED_CMD='sed -i "" '
+else
+  SED_CMD='sed -i '
 fi
 
-docker-compose up $DAEMON_FLAG
+echo "up.sh: On this OSTYPE=$OSTYPE using sed variant $SED_CMD"
```
tzaffi, line 49: This, in my opinion, is the most controversial portion of the script. It would be "better" to run such substitions in a high level language like Python. But I didn't want to introduce this requirement to SDK's which may not already have a python dependency. I'm also concerned about other OS'es now that I see that `sed` behaves differently on mac OS vs. linux.

TODO: test the functionality on Windows.

scripts/up.sh
tzaffi: @michaeldiamant

scripts/up.sh
tzaffi: ^^^ @michaeldiamant

scripts/up.sh
```diff
@@ -1,79 +1,79 @@
...
+echo "up.sh: apply regex replacements on SANDBOX_CFG=$SANDBOX_CFG"
+$SED_CMD "s|#ALGOD_CHANNEL|$ALGOD_CHANNEL|g" "$SANDBOX_CFG"
```
algochoi, line 51: These sed commands are creating a new file called `_config.harness""` on my machine - I don't know enough to figure out why though

README.md
winder: I believe Indexer has proper unit tests to verify the encoding for these fields now. So removing it is fine by me.

README.md
winder: I think the cucumber unit tests cover all of these things, so I'm fine with removing them. They tested Indexer more than anything else, but that doesn't help find regressions since the versions are pinned.

scripts/up.sh
algochoi: Oh I guess I might be on `Darwin` - I notice that there are double quotes after `sed -i` depending on the OS

.env
winder: This comment reminded me that `goal network create` has a new `--devMode` flag. It allows you to take an existing single-node network template, and turn on dev mode. Setting `DEV_MODE=true` instead of a magic template file might be more natural.

scripts/up.sh
winder: I suspect these options mainly for when I was developing these fixtures. As long as it still works, I think you should go ahead and make the scripts work for you.

scripts/up.sh
winder: [Sandbox uses substitution like this as well, so I don't think it's a problem.](https://github.com/algorand/sandbox/blob/c2fb76158ffb7830c592403ae4c05671b22516cf/images/algod/Dockerfile#L42)

Would it help to run `sed` with the `--posix` flag (is that available on mac?).

Alternatively, maybe you could use the `envsubset` command?
https://github.com/winder/docker/pull/1#discussion_r925673551

tzaffi: Great tips. I'll check them out.

scripts/up.sh
tzaffi: yeah... maybe that's how the mac os hack works. I'll be investigating some ideas that Will put up and look into improving the way substitution is working.

.env
```diff
@@ -0,0 +1,31 @@
...
+ALGOD_CHANNEL="nightly"
+
+# Used when TYPE==source:
+ALGOD_URL="https://github.com/algorand/go-algorand"
+ALGOD_BRANCH="master"
+ALGOD_SHA=""
+
+# Used regardless of TYPE:
+NETWORK_TEMPLATE="images/algod/DevModeNetwork.json"  # refers to the ./images directory in the sandbox repo:
+NETWORK_NUM_ROUNDS=3000
```
tzaffi, line 13:
```suggestion
NETWORK_NUM_ROUNDS=30000
```

scripts/up.sh
tzaffi: I've replaced `sed` with `envsubst` so there should be less crud. Please try again.

scripts/up.sh
tzaffi: `envsubst` works, thank you!

.env
```diff
@@ -0,0 +1,30 @@
...
+# Used when TYPE==channel:
+ALGOD_CHANNEL="nightly"
+
+# Used when TYPE==source:
+ALGOD_URL="https://github.com/algorand/go-algorand"
+ALGOD_BRANCH="master"
+ALGOD_SHA=""
+
+# Used regardless of TYPE:
+NETWORK_TEMPLATE="images/algod/DevModeNetwork.json"  # refers to the ./images directory in the sandbox repo:
```
tzaffi, line 12:
```suggestion
NETWORK_TEMPLATE="images/algod/DevModeNetwork.json"  # refers to the ./images directory in the sandbox repo
```

.env
```diff
@@ -0,0 +1,30 @@
...
+# Used regardless of TYPE:
+NETWORK_TEMPLATE="images/algod/DevModeNetwork.json"  # refers to the ./images directory in the sandbox repo
+NETWORK_NUM_ROUNDS=30000
+NODE_ARCHIVAL="False"
+INDEXER_BRANCH="develop"
+INDEXER_SHA=""
+
+# Sandbox configuration:
+SANDBOX_URL="https://github.com/algorand/sandbox"
+SANDBOX_BRANCH="configurable-ports"
```
tzaffi, line 20: ^^^ TODO: revert to master once `configurable-ports` merged

config.harness
```diff
@@ -0,0 +1,15 @@
+export ALGOD_CHANNEL="$ALGOD_CHANNEL"
```
tzaffi, line 1: env vars are substituted in `up.sh` before starting the sandbox with `/sandbox up harness`

features/integration/algod.feature
```diff
@@ -20,8 +20,8 @@ Feature: Algod
     Given a kmd client
     And wallet information
     Then I get transactions by address and round
-    And I get transactions by address only
-    And I get transactions by address and date
+    # And I get transactions by address only
+    # And I get transactions by address and date
```
tzaffi, line 24: requiring indexer v1

features/integration/algod.feature
```diff
@@ -30,19 +30,19 @@ Feature: Algod
     When I get the private key
     And I sign the transaction with the private key
     And I send the transaction
-    Then I can get the transaction by ID
+    # Then I can get the transaction by ID
```
tzaffi, line 33: indexer v1 ^^^

features/integration/algod.feature
```diff
@@ -30,19 +30,19 @@ Feature: Algod
...
-      | 0   |
-      | 1   |
+  # Scenario Outline: Get Transactions By Address and Limit Count
+  #   Given a kmd client
+  #   And wallet information
+  #   When I get recent transactions, limited by <cnt> transactions
+  #   Examples:
+  #     | cnt |
+  #     | 0   |
+  #     | 1   |
```
tzaffi, line 45: ^^^ indexer v1

scripts/up.sh
```diff
@@ -1,79 +1,52 @@
...
-    i) unset DAEMON_FLAG; ;;
-    s) SKIP_BUILD=1; ;;
-    t) TYPE_OVERRIDE=$OPTARG; ;;
-    h) show_help; exit 0 ;;
-    p) unset PARALLEL; ;; 
-  esac
-done
+# Make sure test-sdk sandbox isn't running and clean up any docker detritous
+echo "up.sh: Before bootrapping, try cleaning up first..."
+./scripts/down.sh -f "$ENV_FILE"
```
tzaffi, line 19:
```suggestion
./scripts/down.sh
```

scripts/up.sh
```diff
@@ -1,79 +1,52 @@
...
-    f) ENV_FILE=$OPTARG; ;;
-    i) unset DAEMON_FLAG; ;;
-    s) SKIP_BUILD=1; ;;
-    t) TYPE_OVERRIDE=$OPTARG; ;;
-    h) show_help; exit 0 ;;
-    p) unset PARALLEL; ;; 
-  esac
-done
+# Make sure test-sdk sandbox isn't running and clean up any docker detritous
+echo "up.sh: Before bootrapping, try cleaning up first..."
```
tzaffi, line 18:
```suggestion
echo "up.sh: Before bootstrapping, try cleaning up first..."
```

README.md
```diff
@@ -72,9 +73,9 @@ However, a few are not fully supported:
 
 | tag                             | SDK's which implement        |
 | ------------------------------- | ---------------------------- |
-| @application.evaldelta          | Java only                    |
```
tzaffi, line 75: How about the remove of `@application.evaldelta` - not running since February. @winder @jasonpaulos @ahangsu

scripts/up.sh
```diff
@@ -1,79 +1,46 @@
...
 
-# Defaults
-TYPE_OVERRIDE=""
-DAEMON_FLAG="-d"
-SKIP_BUILD=0
-SHOW_HELP=0
-ENV_FILE=".up-env"
-PARALLEL="--parallel"
-
-show_help() {
```
tzaffi, line 18: removing interactivity of `up.sh`

README.md
winder: Is this different than here? https://github.com/algorand/algorand-sdk-testing/pull/220#discussion_r939797966

I think it can be removed.

tzaffi: Nahhh.. Sorry, I was reading your answer without re-reading my question. Missing caffeine. Thx!

.env
tzaffi: I'll leave this for a future improvement. This requires some further investigation for how to deal with template file in the sandbox.

.env
```diff
@@ -0,0 +1,32 @@
+# Used to determine sandbox build type:
```
michaeldiamant, line 1: @tzaffi Unrelated to the line:  From PR description, responding to _ Get consensus on the best way to move forward with this PR._

> Medium - Continue with the current PR and leave the V1 Endpoint Problem to a future PR. The current approach requires the step removal/modifications described below.

Having made a brief pass through PRs, the quoted approach makes sense to me.

scripts/up.sh
```diff
@@ -1,79 +1,54 @@
...
+
+echo ""
+echo "up.sh: seconds it took to get to end of cloning sandbox into $LOCAL_SANDBOX_DIR: " + $(($(date "+%s") - $START))
+echo ""
+echo "up.sh: bringing up network with TYPE=$TYPE configuration."
+
+pushd "$LOCAL_SANDBOX_DIR"
 
-docker-compose up $DAEMON_FLAG
+./sandbox up -v harness
```
michaeldiamant, line 53: @tzaffi Optional:  Can you either remove (preferred - less effort) or make `-v` a user-configurable flag (default to off)?

* During my local tests, verbose mode output swamps all other output.  I think the user is susceptible to missing sandbox-related output.
* If wiring through a flag is more effort than you prefer, then I think it's reasonable to omit the flag entirely.  A user can make local edits to enable the flag (as I did to disable it).
* If you strongly prefer as is, then I'm good to resolve as is.  I don't feel strongly enough to hold the PR.

tzaffi: I believe it's easy enough to wire it through and default to silent. I'll keep this issue open until I'm clear of the outcome.

scripts/up.sh
```diff
@@ -1,79 +1,54 @@
...
 else
   echo "Unknown environment: $TYPE"
   exit 1
 fi
 
-echo "Bringing up network with '$TYPE' configuration."
-
-# Make sure it isn't running
+# Make sure test-sdk sandbox isn't running and clean up any docker detritous
+echo "up.sh: Before bootstrapping, try cleaning up first..."
```
michaeldiamant, line 26: @tzaffi Optional:  Rather than inline definition of script name (e.g. `up.sh`), propose retrieving script name like https://stackoverflow.com/a/192337 (or similar).

My concern is the log messages will be stale after a script rename.  I think a rename is unlikely and I don't feel strongly about the feedback.

scripts/up.sh
```diff
@@ -1,79 +1,54 @@
...
-docker-compose rm --force
 
-# When developing, it's often useful to skip the build phase.
-if [[ $SKIP_BUILD -eq 0 ]]; then
-  docker-compose build --no-cache $PARALLEL
-fi
+rootdir=$(dirname "$0")
+pushd "$rootdir"/.. > /dev/null || exit
+
+git clone --branch "$SANDBOX_BRANCH" --single-branch "$SANDBOX_URL" "$LOCAL_SANDBOX_DIR"
```
michaeldiamant, line 35: @tzaffi Minor and optional:  To speed up git clone operations, it's possible to clone _only_ the most recent commit (source: https://stackoverflow.com/a/1210012).

sandbox is probably a smaller repo, so treat optionally.  If nothing else, I see a benefit because the clone statement is liable to be copy-pasted later on.

```suggestion
git clone --depth 1 --branch "$SANDBOX_BRANCH" --single-branch "$SANDBOX_URL" "$LOCAL_SANDBOX_DIR"
```

scripts/up.sh
```diff
@@ -1,79 +1,54 @@
...
-  exit
-fi
-
-# Load environment.
-source $ENV_FILE
-
-# Choose which dockerfile to use.
-TYPE=${TYPE_OVERRIDE:-$TYPE}
 if [[ $TYPE == "channel" ]] || [[ $TYPE == "source" ]]; then
   export TYPE="$TYPE"
```
michaeldiamant, line 12: @tzaffi Is the highlighted export necessary?  

Since `TYPE` is sourced from `.env` + exported (due to `set -a`), I think the statement can be safely removed.  I confirmed locally that sandbox still starts up with the line removed. 

```suggestion

```

scripts/up.sh
```diff
@@ -1,79 +1,54 @@
...
-# Defaults
-TYPE_OVERRIDE=""
-DAEMON_FLAG="-d"
-SKIP_BUILD=0
-SHOW_HELP=0
-ENV_FILE=".up-env"
-PARALLEL="--parallel"
+ENV_FILE=".env"
+echo "up.sh: sourcing environment vars from-->$ENV_FILE"
+source "$ENV_FILE"
```
michaeldiamant, line 9: @tzaffi Github won't allow me to add a _suggestion_ here.  Have you considered narrowing the scope of `set -a` to be the following?

```
set -a
source "$ENV_FILE"
set +a
```

* As shown, the PR casts a wide net.  It exports _all_ variables in `up.sh`.  From my review, it seems only the highlighted line must be exported.
* I tried the suggestion locally + removing `set -a` above and sandbox starts.

scripts/up.sh
```diff
@@ -1,79 +1,54 @@
 #!/usr/bin/env bash
-#
-# Bring up the SDK test environment.
 
-set -e
+set -ae
```
michaeldiamant, line 3: @tzaffi Optional:   Since I reviewed the shell configuration for exports, I optionally _suggest_ the configuration I default to (as discussed in https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/).

* Treat optionally - I don't feel strongly.  I leave the suggestion because the configuration has helped me track down errors.
* I deliberately omit `-x`.  I don't think it's needed by default.

```suggestion
set -euo pipefail
```

scripts/up.sh
```diff
@@ -1,79 +1,54 @@
...
+echo ""
+echo "up.sh: seconds it took to get to end of cloning sandbox into $LOCAL_SANDBOX_DIR: " + $(($(date "+%s") - $START))
+echo ""
+echo "up.sh: bringing up network with TYPE=$TYPE configuration."
+
+pushd "$LOCAL_SANDBOX_DIR"
 
-docker-compose up $DAEMON_FLAG
+./sandbox up -v harness
+echo "up.sh: seconds it took to finish with harness ($LOCAL_SANDBOX_DIR) up and running: " + $(($(date "+%s") - $START))
```
michaeldiamant, line 54: @tzaffi A few minor remarks here:

* If you haven't already, can you run scripts through shellcheck?  The change from `$START` to `START` comes from shellcheck and ought to be applied throughout the PR.
* Optionally, the _suggestion_ removes the `+`.  It's unclear to me if you intended to print the `+` or if it's intended for concatenation.  If the latter, it's not needed.
* Optionally, the _suggestion_ attaches the unit of time to the print statement.  Feel welcomed to disregard if you prefer as is.

```suggestion
echo "up.sh: seconds it took to finish with harness ($LOCAL_SANDBOX_DIR) up and running: $(($(date "+%s") - START))s"
```

features/integration/algod.feature
michaeldiamant: @tzaffi Is it your preference to comment out instead of delete the indexer v1 features?

I'm willing to be convinced, but my default is to suggest deleting it.  My thinking is future maintainers will have difficulty understanding _why_ these lines are commented out.

scripts/up.sh
tzaffi: nice!

scripts/up.sh
tzaffi: I'll double check!

scripts/up.sh
tzaffi: I'll double check.

scripts/up.sh
tzaffi: will try

scripts/up.sh
tzaffi: Another good one to try.

README.md
```diff
@@ -35,11 +39,11 @@ These reside in the [unit features directory](features/unit)
...
 | @unit.responses.messagepack       | REST Client MessagePack Unit Tests                    |
 | @unit.responses.messagepack.231   | REST Client MessagePack Unit Tests for Indexer 2.3.1+ |
+| @unit.sourcemap                    | Test the sourcemap endpoint.                        |
 | @unit.tealsign                    | Test TEAL signature utilities.                        |
 | @unit.transactions                | Transaction encoding tests.                           |
 | @unit.transactions.keyreg         | Keyreg encoding tests.                                |
 | @unit.transactions.payment        | Payment encoding tests.                               |
-| @unit.dryrun.trace.application    | DryrunResult formatting tests.                        |
 
 ### Integration Tests
```
michaeldiamant, line 48: @tzaffi Leaving as a visual reminder:  I think the following README sections ought to be updated to reflect the PR.
* https://github.com/algorand/algorand-sdk-testing/#running-tests
* https://github.com/algorand/algorand-sdk-testing/#integration-test-environment
* https://github.com/algorand/algorand-sdk-testing/#start-the-test-environment

If it's possible to consolidate or slim down content, I'm all for it.  My primary concern is to minimize stale information.

Since the PR already covers a lot of ground, if you'd prefer to address the README in a follow up PR, I'm on-board.  I'll follow your lead.

.env
tzaffi: @winder 

I've created this sandbox issue: https://github.com/algorand/sandbox/issues/139

features/integration/algod.feature
tzaffi, line 24:
```suggestion
```

features/integration/algod.feature
tzaffi, line 33:
```suggestion
```

features/integration/algod.feature
tzaffi, line 45:
```suggestion
```

scripts/up.sh
tzaffi: I'm making this configurable by introducing a `VERBOSE_HARNESS` variable in `.env`

scripts/up.sh
tzaffi: it works

README.md
tzaffi: these have been updated

README.md
```diff
@@ -186,12 +188,15 @@ Once the test environment is running you can use `make unit` and `make integrati
...
 
 ![Integration Test Environment](docs/SDK%20Test%20Environment.png)
 
-### Start the test environment
+### Managing the test environment
+
+[up.sh](scripts/up.sh) is used to bring up the test environment. Not surprisingly, [down.sh](scripts/down.sh) brings it all down.
 
-There are a number of [scripts](scripts/) to help with managing the test environment. The names should help you understand what they do, but to get started simply run **up.sh** to bring up a new environment, and **down.sh** to shut it down.
+When starting the environment we default using `go-algorand`'s nightly build. If you're interested in running tests against a specific branch of `go-algorand`, you should set `TYPE="source"` in `.env`
```
michaeldiamant, line 199: @tzaffi Minor _suggestion_:

```suggestion
When starting the environment, we default to using `go-algorand`'s nightly build. If you're interested in running tests against a specific branch of `go-algorand`, you should set `TYPE="source"` in `.env`
```

features/integration/compile.feature
```diff
@@ -27,5 +27,5 @@ Feature: Compile
     When I compile a teal program <teal> with mapping enabled
     Then the resulting source map is the same as the json <sourcemap>
     Examples:
-      | teal                  | sourcemap                                    |
-      | "programs/quine.teal" | "v2algodclient_responsejsons/sourcemap.json" |
\ No newline at end of file
+      | teal                  | sourcemap                                     |
+      | "programs/quine.teal" | "v2algodclient_responsejsons/sourcemap2.json" |
```
michaeldiamant, line 31: @tzaffi For my understanding:  Can you share what motivated the `sourcemap2.json` introduction?

scripts/up.sh
```diff
@@ -1,79 +1,61 @@
...
 
-# Load environment.
-source $ENV_FILE
+set -euo pipefail
+
+START=$(date "+%s")
+
+THIS=$(basename "$0")
+ENV_FILE=".env"
+echo "$THIS: sourcing environment vars from-->$ENV_FILE"
```
michaeldiamant, line 9: @tzaffi Optional - I'd consider printing the absolute path to `$ENV_FILE` to aid debugging.

For example, locally I enabled config overwrite and made no other change in order to see the error message.

```
❯ make harness
./test-harness.sh
~/dev/py-algorand-sdk ~/dev/py-algorand-sdk
~/dev/py-algorand-sdk/test-harness ~/dev/py-algorand-sdk ~/dev/py-algorand-sdk
down.sh: sourcing environment vars from-->.env
down.sh: looking to clean up inside-->
down.sh: directory  does not exist - NOOP
~/dev/py-algorand-sdk ~/dev/py-algorand-sdk
Cloning into 'test-harness'...
remote: Enumerating objects: 254, done.
remote: Counting objects: 100% (254/254), done.
remote: Compressing objects: 100% (218/218), done.
remote: Total 254 (delta 42), reused 149 (delta 19), pack-reused 0
Receiving objects: 100% (254/254), 4.70 MiB | 10.05 MiB/s, done.
Resolving deltas: 100% (42/42), done.
test-harness.sh: OVERWRITE downloaded test-harness/.env with .test-env:
test-harness.sh: seconds it took to get to end of cloning and copying: 1s
~/dev/py-algorand-sdk/test-harness ~/dev/py-algorand-sdk ~/dev/py-algorand-sdk
up.sh: sourcing environment vars from-->.env
./scripts/up.sh: line 15: TYPE: unbound variable
make: *** [harness] Error 1
```

When faced with the shown output, I think it'd accelerate debugging to enable a direct file path copy+paste to inspect `.env`.  

If you prefer as is, feel welcomed to resolve.

features/integration/compile.feature
tzaffi: Formerly `sourcemap.json` was used by both a unit test (look for the tag `@unit.sourcemap`) as well as an integration test (tag `@compile.sourcemap`). Somehow, the expectations of each of these tests contradict each other (at least in the Python SDK). When I modified the integration test to use the new fixture, I was able to get it the tests to pass, but I haven't looked closely at what is the exact problem. I was planning to reach out to @barnjamin as I believe this is related to #353.

michaeldiamant: @tzaffi Got it - thanks for the explanation.

Seems a bit discomforting, but not enough to hold the PR.  Resolving.

michaeldiamant (approved):
@tzaffi Thank you for your investigative work + multiple cycles to bring a net improvement in testing experience + build times!

tzaffi:
> @tzaffi Thank you for your investigative work + multiple cycles to bring a net improvement in testing experience + build times!

Thanks for approving. I will not merge this PR unitll all the "TODO" checkboxes in the PR description have been checked - indicating that all the dependencies are in place for merging.

winder (approved):
Great! Thanks for digging into this and figuring out how many deprecated/duplicate tests we have.

Do you have any idea how many steps this removes?

I found this plugin for Java, maybe the other languages have something similar: https://blog.jdriven.com/2019/01/cucumber-jvm-plugin/

tzaffi:
> Great! Thanks for digging into this and figuring out how many deprecated/duplicate tests we have.
> 
> Do you have any idea how many steps this removes?
> 
> I found this plugin for Java, maybe the other languages have something similar: https://blog.jdriven.com/2019/01/cucumber-jvm-plugin/

Currently, there are [48 steps](https://github.com/algorand/algorand-sdk-testing/pull/220) (see **"Steps Removal Guide"**). @algochoi - 26 thru 48 are the new ones which I identified after I approved the PR. Most helpful was actually the go-sdk which organized the indexer feature step all in one place. Next up, I'll be doing Java and will try to use the tool you linked (now I wish I had started with Java first). I'll update the "Steps Removal Guide" with any more deletions that I discover.

features/integration/compile.feature
```diff
@@ -4,28 +4,28 @@ Feature: Compile
 
   @compile
   Scenario Outline: Compile programs
-    When I compile a teal program <program>
```
tzaffi, line 7: Making variable templating more compliant with SDK expectations.

michaeldiamant (approved):
@tzaffi Thanks for taking in several rounds of feedback - Looking forward to merging!

michaeldiamant:
For future readers, here's the downstream PRs:
* https://github.com/algorand/go-algorand-sdk/pull/360
* https://github.com/algorand/java-algorand-sdk/pull/363
* https://github.com/algorand/js-algorand-sdk/pull/623
* https://github.com/algorand/py-algorand-sdk/pull/367

PR #224, opened 2022-08-23
Bugfix: channel v source

tzaffi:
The sandbox is always defaulting to `release` and the generated `config.harness` isn't sticking because the verbose flag is provided in the wrong order. 

Additionally, after refactoring the SDK Sandboxization PR's to no longer overwrite `.env`, there is no way to pass through the verbosity, except by overwriting. The PR re-enables passing through by also reading `VERBOSE_HARNESS` as a flag (either `-v` or `--verbose`).

For more details regarding "before" and "after" behavior on Circle, see the companion [Python SDK PR](https://github.com/algorand/py-algorand-sdk/pull/373)


scripts/up.sh
```diff
@@ -12,6 +12,17 @@ set -a
...
+while (( "$#" )); do
+  case "$1" in
+    -v|--verbose)
+      VERBOSE_HARNESS=1
+      ;;
+  esac
+  shift
+done
+
+echo "$THIS: VERBOSE_HARNESS=$VERBOSE_HARNESS"
```
tzaffi, line 25:
```suggestion
echo "$THIS: VERBOSE_HARNESS=$VERBOSE_HARNESS"

```

features/integration/compile.feature
```diff
@@ -27,5 +27,5 @@ Feature: Compile
     When I compile a teal program "<teal>" with mapping enabled
     Then the resulting source map is the same as the json "<sourcemap>"
     Examples:
-      | teal                | sourcemap                                   |
-      | programs/quine.teal | v2algodclient_responsejsons/sourcemap2.json |
\ No newline at end of file
+      | teal                | sourcemap                                  |
```
tzaffi, line 30: Not sure why this is failing again. Might be a flaky test that depends on the order run in comparison with other features/steps.

algochoi (commented):
The changes seem to work properly on my local machine! what is the problem with the `compile.feature` file?

tzaffi:
> The changes seem to work properly on my local machine! what is the problem with the `compile.feature` file?

@algochoi - I created issue #226 and commit to spending some time in the next couple of days trying to understand what is actually going on. Are you good on other fronts?

features/integration/compile.feature
ahangsu: just a dumb question, maybe I don't know what's behind the curtain but, `sourcemap.json` and `sourcemap2.json` are not the same, why replacing to a different `sourcemap.json` for a same teal file solves some problem?

tzaffi: Originally, before I worked on sandboxization, they were the same. Then it started failing, and I was getting different results for the integration test. Looking at the results, I [crafted a new `sourcemap2.json`](https://github.com/algorand/algorand-sdk-testing/pull/220/commits/63f0947116e87a6afa299c9f36698c8170f90db7) and got the tests to pass. Now, the bug seems to have gone away. Maybe it has to do with different behavior against `go-algorand`'s `master` v. `feature/avm-box`. I'm not sure and that's what I'm going to investigate in #226.

ahangsu: yea, that matches with my hunch too, that something inside go-algorand produces different sourcemap behavior, and thus the old `sourcemap.json` fail to match expected results.

scripts/up.sh
```diff
@@ -12,6 +12,17 @@ set -a
...
 
+# TODO: allow more than just --verbose env var overrides
+while (( "$#" )); do
+  case "$1" in
+    -v|--verbose)
+      VERBOSE_HARNESS=1
+      ;;
+  esac
+  shift
+done
```
ahangsu, line 23: so I assume that, this chunk of shell actually can read the whole arg option, and find `-v/--verbose` and assign `VERBOSE_HARNESS` to be 1, right? looks neat.

tzaffi: thx - but just copy pasta from other scripts we have

features/integration/compile.feature
algochoi: PR to fix this: https://github.com/algorand/algorand-sdk-testing/pull/227

I'm ready to approve this PR after the above one is merged

algochoi (approved):
Thank you for the fix! Can confirm the builds now work 😄

features/integration/compile.feature
```diff
@@ -28,4 +28,4 @@ Feature: Compile
     Then the resulting source map is the same as the json "<sourcemap>"
     Examples:
       | teal                | sourcemap                                   |
-      | programs/quine.teal | v2algodclient_responsejsons/sourcemap.json  |
\ No newline at end of file
+      | programs/quine.teal | v2algodclient_responsejsons/sourcemap2.json |
```
ahangsu, line 31: nit: is this correct? sourcemap2.json is deleted now

tzaffi: annoying merge conflict ... thx for the catch

ahangsu (approved):

features/integration/compile.feature
ahangsu: wait, still not reverted...

ahangsu (approved):
really approve

PR #232, opened 2022-09-06
Integration Test for Box endpoints in Indexer

ahangsu:


features/integration/applications.feature
```diff
@@ -4,6 +4,7 @@ Feature: Applications
       And a kmd client
       And wallet information
       And an algod v2 client connected to "localhost" port 60000 with token "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+      And an indexer v2 client
```
tzaffi, line 7: Yay. Bringing back the indexer.

utils/unused_steps_analysis.ipynb
```diff
@@ -22,6 +22,7 @@
     "\n",
     "%pip install gherkin-official\n",
     "%pip install fuzzywuzzy\n",
+    "%pip install pandas\n",
```
tzaffi, line 25: 👍

features/integration/applications.feature
```diff
@@ -125,15 +126,79 @@ Feature: Applications
...
+      And according to "indexer", by parameter max 6, the current application should have 4 boxes.
+
+      # delete one box
+      Then I build an application transaction with the transient account, the current application, suggested params, operation "call", approval-program "", clear-program "", global-bytes 0, global-ints 0, local-bytes 0, local-ints 0, app-args "str:delete,str:name", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 0, boxes "0,str:name"
+      And I sign and submit the transaction, saving the txid. If there is an error it is "".
+      And I wait for the transaction to be confirmed.
+      And according to "algod", the current application should have the following boxes "str:foo bar,b64:APj/IA==,b64:MTE0NTE0".
+
+      # move to indexer testing steps
+      And I sleep for 1 seconds for indexer to digest things down.
```
tzaffi, line 198: 🤣

tzaffi (commented):
I put forth some minor adjustments. The most important ones are to expect box name and values to be b64 encoded (rather than arbitrary abi), and separating results by `:` as opposed to `,`. Less imporant are testing indexer's `next` and unifying the steps.

features/integration/applications.feature
ahangsu: well, one finding I realized during the testcase writing is that, *a certain wait interval* is needed, for no waiting leads to indexer fail to reflect changes applied on algod side (e.g., deletion).

But indeed, the interval may be even shorter than 1 sec. My thinking is: this might be good nuf? For I am concerned that testing will be flaky if we be aggressive and trim down even more... I confess that this is a quick dirty trick.

tzaffi: Given your comment, I recommend that you change the step to `I sleep for {int} milliseconds for indexer to digest things down.`

This makes it a bit more future proof. We can start with a small number like 500 and then increase if we find tests are failing.

features/integration/applications.feature
```diff
@@ -149,17 +149,17 @@ Feature: Applications
       And I fund the current application's address with 100000000 microalgos.
       Then I get the account address for the current application and see that it matches the app id's hash
 
-      # create a box called "str:name"
-      And I build an application transaction with the transient account, the current application, suggested params, operation "call", approval-program "", clear-program "", global-bytes 0, global-ints 0, local-bytes 0, local-ints 0, app-args "str:create,str:name", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 0, boxes "0,str:name"
+      # create a box called "str:name" (i.e., b64:bmFtZQ==)
+      And I build an application transaction with the transient account, the current application, suggested params, operation "call", approval-program "", clear-program "", global-bytes 0, global-ints 0, local-bytes 0, local-ints 0, app-args "str:create,b64:bmFtZQ==", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 0, boxes "0,b64:bmFtZQ=="
```
tzaffi, line 153: I think it's totally ok to let the `app-args` be abi-encoded. I'm only asking for changes in the _parsed response_ assertions for box names and values.

features/integration/applications.feature
```diff
@@ -125,15 +126,79 @@ Feature: Applications
...
+
+      # delete one box
+      Then I build an application transaction with the transient account, the current application, suggested params, operation "call", approval-program "", clear-program "", global-bytes 0, global-ints 0, local-bytes 0, local-ints 0, app-args "str:delete,str:name", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 0, boxes "0,str:name"
+      And I sign and submit the transaction, saving the txid. If there is an error it is "".
+      And I wait for the transaction to be confirmed.
+      And according to "algod", the current application should have the following boxes "str:foo bar,b64:APj/IA==,b64:MTE0NTE0".
+
+      # move to indexer testing steps
+      And I sleep for 1 seconds for indexer to digest things down.
+      And according to "indexer", the contents of the box with name "b64:APj/IA==" in the current application should be "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA". If there is an error it is "".
```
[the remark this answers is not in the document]
ahangsu: the indexer query for non-empty box is now refleced on line 182, I will try to add a query for non-existing box to trigger some error

features/integration/applications.feature
```diff
@@ -125,15 +126,79 @@ Feature: Applications
...
+      And I sign and submit the transaction, saving the txid. If there is an error it is "".
+      And I wait for the transaction to be confirmed.
+      Then according to "algod", the contents of the box with name "str:foo bar" in the current application should be "YmF6IHF1eAAAAAAAAAAAAAAAAAAAAAAA". If there is an error it is "".
+
+      # full check confirmed by both algod and indexer
+      Then according to "algod", the current application should have the following boxes "str:foo bar,b64:APj/IA==,str:name,b64:MTE0NTE0".
+      And according to "algod", by parameter max 2, the current application should have 2 boxes.
+      And according to "algod", by parameter max 6, the current application should have 4 boxes.
+      And I sleep for 1 seconds for indexer to digest things down.
+      And according to "indexer", the current application should have the following boxes "str:foo bar,b64:APj/IA==,str:name,b64:MTE0NTE0".
```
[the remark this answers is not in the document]
ahangsu: a few things to correct and a few things to adapt:
- `str:foo bar` is not abi encoding, it is encoded in just string, right?
- I am changing all the step expectations to be b64 encoded (with b64 prefix).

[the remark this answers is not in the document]
ahangsu: I would like to push a little back on the part of `returned in order`, for this would be very strongly related with the order of operations. To couple with parameter `max`/`limit`, one can hardly tell what would be returned under algod.

These are quite implicit and hidden, such that I would not recommend to require these in step for algod.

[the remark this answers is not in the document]
tzaffi: Right, I should use the nomenclature _app-arg encoded_ which happens to now allow abi-encoding in the case of boxes.

I'm resolving.

.env
```diff
@@ -18,7 +18,7 @@ NETWORK_NUM_ROUNDS=30000
 INDEXER_URL="https://github.com/algorand/indexer"
 NODE_ARCHIVAL="False"
 INDEXER_BRANCH="tzaffi/box-search"
-INDEXER_SHA=""
+INDEXER_SHA="7e7f849dd51ca41d0a8293bcf72f478726a3e036"
```
tzaffi, line 21: nice workaround!

features/integration/applications.feature
```diff
@@ -125,15 +126,83 @@ Feature: Applications
...
+
+      # move to indexer testing steps
+      And I sleep for 500 milliseconds for indexer to digest things down.
+      And according to "indexer", the contents of the box with name "str:name" in the current application should be "". If there is an error it is "no application boxes found".
+      And according to "indexer", the contents of the box with name "b64:APj/IA==" in the current application should be "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA". If there is an error it is "".
+      And according to "indexer", the current application should have the following boxes "Zm9vIGJhcg==:APj/IA==:MTE0NTE0".
+      And according to "indexer", by parameter max 2, the current application should have 2 boxes.
+      And according to "indexer", by parameter max 0, the current application should have 3 boxes.
+      And according to indexer, by parameter max 1 and next "b64:APj/IA==", the current application should have the following boxes "MTE0NTE0".
+
```
tzaffi, line 208:
```suggestion
```

features/integration/applications.feature
```diff
@@ -125,15 +126,83 @@ Feature: Applications
...
+      And I sign and submit the transaction, saving the txid. If there is an error it is "".
+      And I wait for the transaction to be confirmed.
+      And according to "algod", the current application should have the following boxes "Zm9vIGJhcg==:APj/IA==:MTE0NTE0".
+
+      # move to indexer testing steps
+      And I sleep for 500 milliseconds for indexer to digest things down.
+      And according to "indexer", the contents of the box with name "str:name" in the current application should be "". If there is an error it is "no application boxes found".
+      And according to "indexer", the contents of the box with name "b64:APj/IA==" in the current application should be "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA". If there is an error it is "".
+      And according to "indexer", the current application should have the following boxes "Zm9vIGJhcg==:APj/IA==:MTE0NTE0".
+      And according to "indexer", by parameter max 2, the current application should have 2 boxes.
```
tzaffi, line 205:
```suggestion
      And according to "indexer", with 2 being the parameter that limits results, the current application should have 2 boxes.
```
The reason for the suggested change is that the "max" terminology makes sense for `algod` but not `indexer`. If the step i intended for use by indexer only, you could just change "max" to "limit".

features/integration/applications.feature
```diff
@@ -125,15 +126,83 @@ Feature: Applications
...
-      Then the current application should have the following boxes "b64:APj/IA==".
+      Then according to "algod", the current application should have the following boxes "APj/IA==".
       # Delete last box and check that empty response is handled correctly
       And I build an application transaction with the transient account, the current application, suggested params, operation "call", approval-program "", clear-program "", global-bytes 0, global-ints 0, local-bytes 0, local-ints 0, app-args "str:delete,b64:APj/IA==", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 0, boxes "0,b64:APj/IA=="
       And I sign and submit the transaction, saving the txid. If there is an error it is "".
       And I wait for the transaction to be confirmed.
-      Then the current application should have the following boxes "".
+      Then according to "algod", the current application should have the following boxes "".
+
+    @applications.boxes_indexer_confirmed
```
tzaffi, line 142: Is it possible to continue using the tag `@applications.boxes` rather than introduce a new tag? Since it looks like we'll able to merge all the box features together, I don't see an advantage to breaking up into separate tags, but I might not be seeing the whole picture.

features/integration/applications.feature
tzaffi: Take this suggestion with a grain of salt...not at all a big deal.

features/integration/applications.feature
```diff
@@ -139,8 +139,7 @@ Feature: Applications
       And I wait for the transaction to be confirmed.
       Then according to "algod", the current application should have the following boxes "".
 
-    @applications.boxes_indexer_confirmed
-    Scenario: Exercise indexer after a slew of box creations
```
tzaffi: You don't need the `@application.boxes` tag here?

.env
ahangsu: NOTE: Revert once indexer PR is in good shape and settled down
```suggestion
INDEXER_SHA=""
```

features/integration/applications.feature
ahangsu: yea definitely, I am merging these scenarios under a same tag. Introducing a new tag is for my lack of knowledge of gherkin.

features/integration/applications.feature
ahangsu: using it now!

features/integration/applications.feature
ahangsu: put this on the top of my stack tomorrow

tzaffi (approved):
Approving as already started approving SDK's and my remaining nit isn't a blocker.

features/integration/applications.feature
ahangsu: just adapted your suggestion.

tzaffi:
> Approving as already started approving SDK's and my remaining nit isn't a blocker.

LEBTM

features/integration/applications.feature
```diff
@@ -125,15 +126,82 @@ Feature: Applications
...
+
+      # move to indexer testing steps
+      And I sleep for 500 milliseconds for indexer to digest things down.
+      And according to "indexer", the contents of the box with name "str:name" in the current application should be "". If there is an error it is "no application boxes found".
+      And according to "indexer", the contents of the box with name "b64:APj/IA==" in the current application should be "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA". If there is an error it is "".
+      And according to "indexer", the current application should have the following boxes "Zm9vIGJhcg==:APj/IA==:MTE0NTE0".
+      And according to "indexer", with 2 being the parameter that limits results, the current application should have 2 boxes.
+      And according to "indexer", with 0 being the parameter that limits results, the current application should have 3 boxes.
+      And according to indexer, with 1 being the parameter that limits results, and "b64:APj/IA==" being the parameter that sets the next result, the current application should have the following boxes "MTE0NTE0".
```
algochoi, line 207: Perhaps we can roll this step up with: `And according to <client>, the current application should have the following boxes "Zm9vIGJhcg==:APj/IA==:MTE0NTE0".`:

We could combine the two and have a step like:
`And according to <client>, with limit <max> and next "<next>", the current application should have the following boxes "<boxes>".` and ignore `next` if it is an `algod` client.

algochoi (commented):
Nice work - also it looks like the indexer tests only add about 20 seconds to the integration test process, but currently adds about 7 minutes to the overall CI build ([before](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/913/workflows/165455d8-ef69-45bb-8b49-a0aed68dea9f/jobs/4159) and [after](https://app.circleci.com/pipelines/github/algorand/py-algorand-sdk/950/workflows/de66b218-2a8b-4a8d-a7a2-7842b6776d26/jobs/4442)). I think it's likely because we are using a custom build, but it might be worth a look to check if there is significant difference after using the `channel` build for indexer.

features/integration/applications.feature
```diff
@@ -125,15 +126,82 @@ Feature: Applications
...
-      Then the current application should have the following boxes "b64:APj/IA==".
+      Then according to "algod", the current application should have the following boxes "APj/IA==".
       # Delete last box and check that empty response is handled correctly
       And I build an application transaction with the transient account, the current application, suggested params, operation "call", approval-program "", clear-program "", global-bytes 0, global-ints 0, local-bytes 0, local-ints 0, app-args "str:delete,b64:APj/IA==", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 0, boxes "0,b64:APj/IA=="
       And I sign and submit the transaction, saving the txid. If there is an error it is "".
       And I wait for the transaction to be confirmed.
-      Then the current application should have the following boxes "".
+      Then according to "algod", the current application should have the following boxes "".
+
+   @applications.boxes
```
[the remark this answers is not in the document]
ahangsu: I guess we are not preserving the convention for `@indexer` tag, by rolling forward with unified approach, since I assume it is easier to test indexer in testing with sandbox.

features/integration/applications.feature
ahangsu: This is also somewhat brought up by Zeph previously in a thread of discussion: https://github.com/algorand/algorand-sdk-testing/pull/232#discussion_r969017995, and my understanding on endpoint *boxes* result returning is that, if we ask with `limit`, we are asking for a subset of all boxes under an app.

A problem directly arise: which part of subset is the endpoint returning, or how can we manipulate the order such that we can get the result for expected boxes? These are implicit and hidden. Without special notifying, the test steps in scenario cannot be easily swapped, which smells error prone and hard to maintain for me.

ahangsu: (I mean, the `limit` case only applies on algod, for indexer it seems returning is ordered? but this is still an obstacle)

ahangsu: yea that makes sense, I can drop a note in testcase to explain the thinking process.

algochoi: thanks! comment looks good so resolving

PR #233, opened 2022-09-09
EZ Integration Tests

tzaffi:
WIP

Related PR's at various stages of WIP'ness:
* [go-algorand](https://github.com/tzaffi/go-algorand/pull/7)
  * Its [ezTraces_test.go](https://github.com/tzaffi/go-algorand/blob/8347a98b85f6d2f59ecb4fe1e5a0def8a9cc8b6a/test/e2e-go/restAPI/ezTraces_test.go#L168) generates the [disassemble.json fixture](https://github.com/tzaffi/go-algorand/blob/8347a98b85f6d2f59ecb4fe1e5a0def8a9cc8b6a/test/e2e-go/restAPI/ezTraces/disassemble.json) that is gonna end up in this PR.
* [js-algorand-sdk](https://github.com/algorand/js-algorand-sdk/pull/647)

## Random Thoughts
- [ ] Shouldn't we also automatically test paths?
  - [ ] In the [disassemble example](https://github.com/algorand/algorand-sdk-testing/blob/8767033f3354f2b15af13078207fefd93bd36ff5/features/resources/ezTraces/disassemble.json#L4) there is no difference between path and route. However, all the SDK's support introspecting path via a dedicated method, and the go rest client tracer ought to be able to spot this as well
  - [ ] This may obviate [one sort of unit test](https://github.com/algorand/algorand-sdk-testing/blob/b6cee0c4d95a3cfe20749ed5999df43b7e3472d6/features/unit/v2algodclient_paths.feature#L73-L91) that we always set up.
- [ ] The `POST` portion of any such test, could be understood to be the setup for the READONLY `GET` portion. A test in indexer could pick up where the rest-client left off, and modify the fixture to include indexer queries at the end.
- [ ] There is an SDK transaction creation unit tests with a "golden" containing [the encoded signed transaction to be sent](https://github.com/algorand/algorand-sdk-testing/blob/b6cee0c4d95a3cfe20749ed5999df43b7e3472d6/features/unit/transactions.feature#L18).
  - [ ] Could such a unit test be fixturized as well? Maybe not in an integration test, as absolute determinism may be required to generate the same transaction each time? Or maybe the transaction details could be asserted against instead?


features/integration/ez.feature
```diff
@@ -0,0 +1,17 @@
...
+    * I make a transaction signer for the transient account.
+
+  @ez.disassemble
+  Scenario: 
+    Given an ez request/response array fixture "disassemble.json" to save in "disassemble"'s context.
+    * Parse and save as ordered each "disassemble" request's route, path, verb, content-type, optional params and post-data with encoding info.
+    Then Sanity check that "disassemble"'s request data matches its verbs.
+    * Decode "disassemble"'s data as needed in preparation for making requests.
+    * Parse and save as ordered each "disassemble" response body in the given format.
+    * Infer each SDK method from "disassemble"'s requests and run it live each saving its response.
```
tzaffi, line 16:
```suggestion
    * Infer each SDK method from "disassemble"'s requests and run it live saving its response.
```

PR #192, opened 2022-06-16
Feature: Box Storage

algochoi:
This PR stores all the diffs/commits to support application Boxes. This branch roughly tracks the `feature/avm-box` branch in `go-algorand`.

PR #235, opened 2022-09-20
add clear state teal prog 8

ahangsu:
Teal 8 Clear state program


ahangsu:
My bad, the title should suggest clear state program, so that might be somewhat mislead you. The motivation of this PR stem from the thread of discussion: https://github.com/algorand/algorand-sdk-testing/pull/192#discussion_r974694479

Though the suggestion on moving programs that are serving clear state purpose into another directory is making sense, I would prefer them to be simply put inside `resource/programs`.

tzaffi (approved):
LGTM!

algochoi (approved):
Looks good - I think for maintainability, it would be nice to put these shorter programs in another directory for the future.

PR #245, opened 2022-10-05
REST API Tests: algod block hash, indexer blocks parameter

winder:


README.md
```diff
@@ -22,15 +22,17 @@ These reside in the [unit features directory](features/unit)
 | @unit.abijson                     | ABI types and method encoding/decoding unit tests.    |
 | @unit.algod                       | Algod REST API unit tests.                            |
 | @unit.algod.ledger_refactoring    |                                                       |
+| @unit.algod.blockhash            | Algod REST API unit tests.                            |
```
tzaffi, line 25:
```suggestion
| @unit.algod.blockhash             | Algod REST API unit tests.                            |
```

tzaffi (commented):
Looks good. I can see both sides of the following argument here:

* We should aim to consolidate the new tags as much as possible. Could we get away with introducing a single new tag such as `unit.blocksummary` ?
* We should keep as is because the different tags broadcast different intents.

tzaffi (approved):
LGTM

PR #246, opened 2022-10-31
Merge in develop and update support for go-algorand#4689 

michaeldiamant:
Prepares to merge the parent branch by performing a final merge from `master` and accounts for a behavior change in https://github.com/algorand/go-algorand/pull/4689.


features/integration/applications.feature
```diff
@@ -183,7 +183,6 @@ Feature: Applications
 
       # full check confirmed by both algod and indexer
       Then according to "algod", the current application should have the following boxes "Zm9vIGJhcg==:APj/IA==:bmFtZQ==:MTE0NTE0".
-      And according to "algod", with 2 being the parameter that limits results, the current application should have 2 boxes.
```
michaeldiamant, line 186: Adding a thread to highlight the diff - The removed step no longer reliably works after https://github.com/algorand/go-algorand/pull/4689.  

The test creates > 2 boxes and based on non-deterministic timing, the test _may_ pass if only 2 boxes created when the step runs.

tzaffi: It appears that the previous step passes with 4 boxes. Therefore, the following step, which was passing before the change, should reliably error according to the new functionality (if my understanding is correct). Ideally, we would assert this erroring behavior here so that proper SDK handling of the error case is verified. However, I expect that the necessary mechanisms for handing 400's are already in place and I won't advocate for a new test case here and I see that we are at lease [verifying that the http response is erroring elsewhere](https://github.com/michaeldiamant/go-algorand/blob/fd6215b201505d031f176cdc6fe2339699bd48d9/test/e2e-go/restAPI/restClient_test.go#L1404).

tzaffi (approved):
Approving as in my opinion, enriching the feature to assert an error case for the `boxes` endpoint should not hold up the deployment.

features/integration/applications.feature
michaeldiamant: Thanks for the double check - I also previously confirmed go-algorand integration tests cover the error case.   Otherwise, I'd _at least_ make a change there.

PR #248, opened 2022-11-07
Test app call txn decoding

jasonpaulos:
Add an additional step to the existing app call transaction cases to ensure decoding is also tested.

This was prompted by https://github.com/algorand/js-algorand-sdk/issues/688, which discovered a bug in the JS SDK transaction decoding for box transactions.

SDK status with respect to this change:
- [x] JS: to be fixed by algorand/js-algorand-sdk#690
- [x] Python: to be fixed by https://github.com/algorand/py-algorand-sdk/pull/398
- [x] Go: passing with https://github.com/algorand/go-algorand-sdk/pull/433
- [x] Java: to be fixed by https://github.com/algorand/java-algorand-sdk/pull/422


algochoi (approved):
LGTM 👍

PR #253, opened 2022-12-06
Test ABI reference arguments during creation

jasonpaulos:
This PR adds a new integration test scenario to cover calling an ABI method with reference types during an application creation.

Due to the bug pointed out in https://github.com/algorand/py-algorand-sdk/pull/406, we want increase coverage for that scenario in goal and the SDKs.

As part of this work, I also updated the PyTeal contract at https://gist.github.com/jasonpaulos/99e4f8a75f2fc2ec9b8073c064530359 to use the ABI router, since the prior contract did not perfectly follow ARC-4 routing rules. This caused the changes in `features/resources/programs/abi_method_call.teal` and the corresponding `.teal.tok` file.

I have manually ran the new test on all SDKs, and it passes everywhere except for Python. But with https://github.com/algorand/py-algorand-sdk/pull/406, Python passes.

Note that while this new test adds some coverage for reference types, additional unit tests would be preferable to have higher confidence. At the moment, there is no cucumber step to specify _additional_ reference types to be included in a transaction, outside of those required as arguments.


algochoi (approved):
Thanks for the changes! LGTM - It looks correct on my local run in the Python SDK as well.

PR #257, opened 2022-12-12
REST API Tests: statedelta endpoint.

winder:


Eric-Warehime (approved):

winder:
I removed "min-balance" from the sample responses. This is a temporary workaround to the fact that Indexer/Algod have different account objects.

PR #265, opened 2023-02-24
Simulate endpoint tests

algochoi:
Adds tests for:
* Calling the simulate endpoint for basic payment txns
* Using the simulate method/functionality within an AtomicTransactionComposer
* Check fail-path for inner transaction failures
* Simulating unsigned transactions 

Also changes the following (not-necessarily related to simulate)
* Decrease transient account funding to reduce economic burden on genesis accounts
* Add a method in `RandomByte` TEAL contract to call a bad inner txn

Closes #270 

Related SDK PRs
* https://github.com/algorand/py-algorand-sdk/pull/420
* https://github.com/algorand/js-algorand-sdk/pull/743

### DIFF
--- a/.env
+++ b/.env
@@ -2,14 +2,14 @@
 VERBOSE_HARNESS=0
 
 # Used to determine sandbox build type:
-TYPE="channel" # OR "source"
+TYPE="source" # OR "source"
 
 # Used when TYPE==channel:
 ALGOD_CHANNEL="nightly"
 
 # Used when TYPE==source:
-ALGOD_URL="https://github.com/algorand/go-algorand"
-ALGOD_BRANCH="master"
+ALGOD_URL="https://github.com/jasonpaulos/go-algorand"
+ALGOD_BRANCH="nonexperimental-simulation"
 ALGOD_SHA=""
 
 # Used regardless of TYPE:
--- a/.env
+++ b/.env
@@ -2,7 +2,7 @@
 VERBOSE_HARNESS=0
 
 # Used to determine sandbox build type:
-TYPE="channel" # OR "source"
+TYPE="source" # OR "source"
 
 # Used when TYPE==channel:
 ALGOD_CHANNEL="nightly"
--- a/.env
+++ b/.env
@@ -22,7 +22,7 @@ INDEXER_SHA=""
 
 # Sandbox configuration:
 SANDBOX_URL="https://github.com/algorand/sandbox"
-SANDBOX_BRANCH="master"
+SANDBOX_BRANCH="experimental-api"
 LOCAL_SANDBOX_DIR=".sandbox"
 
 # replacement values for Sandbox's docker-compose:
--- a/.env
+++ b/.env
@@ -2,7 +2,7 @@
 VERBOSE_HARNESS=0
 
 # Used to determine sandbox build type:
-TYPE="source" # OR "source"
+TYPE="channel" # OR "source"
 
 # Used when TYPE==channel:
 ALGOD_CHANNEL="nightly"
@@ -17,7 +17,7 @@ NETWORK_TEMPLATE="images/algod/DevModeNetwork.json"  # refers to the ./images di
 NETWORK_NUM_ROUNDS=30000
 INDEXER_URL="https://github.com/algorand/indexer"
 NODE_ARCHIVAL="False"
-INDEXER_BRANCH="develop"
+INDEXER_BRANCH="master"
 INDEXER_SHA=""
 
 # Sandbox configuration:
--- a/.env
+++ b/.env
@@ -2,14 +2,14 @@
 VERBOSE_HARNESS=0
 
 # Used to determine sandbox build type:
-TYPE="source" # OR "source"
+TYPE="channel" # OR "source"
 
 # Used when TYPE==channel:
 ALGOD_CHANNEL="nightly"
 
 # Used when TYPE==source:
-ALGOD_URL="https://github.com/jasonpaulos/go-algorand"
-ALGOD_BRANCH="nonexperimental-simulation"
+ALGOD_URL="https://github.com/algorand/go-algorand"
+ALGOD_BRANCH="master"
 ALGOD_SHA=""
 
 # Used regardless of TYPE:
@@ -22,7 +22,7 @@ INDEXER_SHA=""
 
 # Sandbox configuration:
 SANDBOX_URL="https://github.com/algorand/sandbox"
-SANDBOX_BRANCH="experimental-api"
+SANDBOX_BRANCH="master"
 LOCAL_SANDBOX_DIR=".sandbox"
 
 # replacement values for Sandbox's docker-compose:
--- a/features/integration/abi.feature
+++ b/features/integration/abi.feature
@@ -5,7 +5,7 @@ Feature: ABI Interaction
     And a kmd client
     And wallet information
     And suggested transaction parameters from the algod v2 client
-    And I create a new transient account and fund it with 100000000 microalgos.
+    And I create a new transient account and fund it with 10000000 microalgos.
     And I make a transaction signer for the transient account.
     And I build an application transaction with the transient account, the current application, suggested params, operation "create", approval-program "programs/abi_method_call.teal.tok", clear-program "programs/one.teal.tok", global-bytes 0, global-ints 0, local-bytes 1, local-ints 0, app-args "", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 0, boxes ""
     And I sign and submit the transaction, saving the txid. If there is an error it is "".
@@ -98,7 +98,7 @@ Feature: ABI Interaction
     And The app should have returned "ABRnb29kYnllIEFsZ29yYW5kIEZhbg==".
 
   Scenario: Method call delete execution
-    Given I create a new transient account and fund it with 100000000 microalgos.
+    Given I create a new transient account and fund it with 10000000 microalgos.
     And I make a transaction signer for the transient account.
     And I build an application transaction with the transient account, the current application, suggested params, operation "create", approval-program "programs/abi_method_call.teal.tok", clear-program "programs/one.teal.tok", global-bytes 0, global-ints 0, local-bytes 1, local-ints 0, app-args "", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 0, boxes ""
     And I sign and submit the transaction, saving the txid. If there is an error it is "".
--- a/features/integration/simulate.feature
+++ b/features/integration/simulate.feature
@@ -1,6 +1,6 @@
 Feature: Simulating transactions
   Background:
-    Given an algod v2 client connected to "localhost" port 4001 with token "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+    Given an algod v2 client
     And a kmd client
     And wallet information
     And suggested transaction parameters from the algod v2 client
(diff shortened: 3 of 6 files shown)


.env
```diff
@@ -2,14 +2,14 @@
...
-TYPE="channel" # OR "source"
+TYPE="source" # OR "source"
 
 # Used when TYPE==channel:
 ALGOD_CHANNEL="nightly"
 
 # Used when TYPE==source:
-ALGOD_URL="https://github.com/algorand/go-algorand"
-ALGOD_BRANCH="master"
+ALGOD_URL="https://github.com/jasonpaulos/go-algorand"
```
algochoi, line 11: TODO: Set this back to master when experimental tags are removed from simulate endpoint

.env
```diff
@@ -22,7 +22,7 @@ INDEXER_SHA=""
 
 # Sandbox configuration:
 SANDBOX_URL="https://github.com/algorand/sandbox"
-SANDBOX_BRANCH="master"
+SANDBOX_BRANCH="experimental-api"
```
algochoi, line 25: TODO: set this back to master (Theoretically, we won't need to support experimental flags on sandbox when the related simulate endpoint PR is merged into master)

barnjamin (approved):

features/integration/simulate.feature
```diff
@@ -0,0 +1,84 @@
...
+    Given a new AtomicTransactionComposer
+    When I build a payment transaction with sender "transient", receiver "transient", amount 100001, close remainder to ""
+    And I create a transaction with signer with the current transaction.
+    And I add the current transaction with signer to the composer.
+    Then I simulate the current transaction group with the composer
+    And the simulation should succeed without any failure message
+    And I clone the composer.
+    When I add the current transaction with signer to the composer.
+    Then I simulate the current transaction group with the composer
+    And the simulation should report a failure at path "1" with message "transaction already in ledger"
```
tzaffi, line 34: I see the following nested response structure:
```go
Result{
		Version:      version,
		LastRound:    lastRound,
		TxnGroups:    []TxnGroupResult{
                               	Txns:           []TxnResult {
                               	                           	Txn:                        transactions.SignedTxnWithAD
                               	                           	MissingSignature: bool
                               	                        }
                               	FailureMessage: string
                               	// FailedAt is the path to the txn that failed inside of this group
                               	FailedAt: TxnPath []uint64
                              },
		WouldSucceed: true,
}
```
Is this path really just the `FailedAt: TxnPath` inside of **group 0**?

tzaffi: And this provokes an uncomfortable question. Should we be future proof and support more than 1 txn group? I have very mixed feelings about it.

tzaffi (commented):
Looking solid. My main question surrounds what to do (if anything) about simulating without signatures. My intuition is that we _should_ support it. But if we do, then we also need to make it easy to send unsigned transactions and also to collect the results and have some method such as "would_have_succeeded_but_for_signatures()` (please find a better name) that easily answers the question suggested by this terrible method name.

features/integration/simulate.feature
```diff
@@ -0,0 +1,84 @@
...
+    And wallet information
+    And suggested transaction parameters from the algod v2 client
+    And I create a new transient account and fund it with 100000000 microalgos.
+    And I make a transaction signer for the transient account.
+
+  @simulate
+  Scenario Outline: Simulating successful payment transactions
+    Given default transaction with parameters <amt> "<note>"
+    When I get the private key
+    And I sign the transaction with the private key
```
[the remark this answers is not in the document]
algochoi: This is a good point. I originally omitted missing signature tests in the SDKs because some of them currently do not allow unsigned transactions to be encoded and sent over to algod at all if they are missing a signature. In the JS SDK, the encoder throws an error if there is a missing field (signature) in the transaction: https://github.com/algorand/js-algorand-sdk/blob/develop/src/encoding/encoding.ts#L46 

I still think simulating transactions without signatures sounds like a valid use-case so we might have to support this some other way. Tagging @barnjamin as well, and I'll also think about this a bit more.

features/integration/simulate.feature
```diff
@@ -0,0 +1,84 @@
+Feature: Simulating transactions
+  Background:
+    Given an algod v2 client
+    And a kmd client
+    And wallet information
+    And suggested transaction parameters from the algod v2 client
+    And I create a new transient account and fund it with 100000000 microalgos.
```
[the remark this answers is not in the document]
algochoi: ah thanks for noticing, I'm going to be more stingy here in the next commit

features/integration/abi.feature
```diff
@@ -5,7 +5,7 @@ Feature: ABI Interaction
     And a kmd client
     And wallet information
     And suggested transaction parameters from the algod v2 client
-    And I create a new transient account and fund it with 100000000 microalgos.
+    And I create a new transient account and fund it with 10000000 microalgos.
```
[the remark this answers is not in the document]
algochoi: Yeah, I think there are more wasteful tests elsewhere, but we recently have been getting more `overspend` errors, so I figured we should start somewhere.

features/integration/simulate.feature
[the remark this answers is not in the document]
barnjamin: see thread here https://github.com/algorand/py-algorand-sdk/pull/420#discussion_r1122379707

maybe add a new EmptySigner

features/integration/simulate.feature
algochoi: > And this provokes an uncomfortable question. Should we be future proof and support more than 1 txn group? I have very mixed feelings about it.

This is a good point, I think we should future-proof it. i.e. `the simulation should report a failure at group "0", path "1" with message "transaction already in ledger"`

But just curious why you had mixed feelings about it?

tzaffi: "Future proofing" is usually another name for "over-engineering". However, if we know for sure that `/simulate` will support multiple transactions in the near/medium term, then it's not over-engineering, but simply prudent.

@bbroder-algo - any thoughts?

features/integration/simulate.feature
```diff
@@ -0,0 +1,84 @@
...
+    And I make a transaction signer for the transient account.
+
+  @simulate
+  Scenario Outline: Simulating successful payment transactions
+    Given default transaction with parameters <amt> "<note>"
+    When I get the private key
+    And I sign the transaction with the private key
+    And I simulate the transaction
+    Then the simulation should succeed without any failure message
```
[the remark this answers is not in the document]
algochoi: > Should we check that this didn't cost anything? (I'm leaning to "NO" even though I asked the question. But the next one, might be more of a "YES")

I think this would be closer to a ledger test (checking if the ledger accounts have not been modified by the simulation)

> Should we check that it fails for an over-spend?

Sure, I can add this test case without affecting any steps in downstream SDKs.

features/integration/simulate.feature
algochoi: I think it will need to support multiple groups eventually - I'm for it

features/integration/simulate.feature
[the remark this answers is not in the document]
algochoi: Addressed in 90fdb79ac98c2340c04d86fc7e6167f9af8b2b25

features/integration/simulate.feature
algochoi: I pushed [dcc0b66](https://github.com/algorand/algorand-sdk-testing/pull/265/commits/dcc0b66ec2b133434b8be3c65fa13d01164aa966) which adds the group num in the step.

features/integration/simulate.feature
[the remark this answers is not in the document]
algochoi: I'll make an issue on this as well (support missing sigs in the SDKs and simulation) - we can address this in a follow-up PR?

features/integration/simulate.feature
```diff
@@ -0,0 +1,84 @@
...
+    And the simulation should succeed without any failure message
+    And I clone the composer.
+    When I add the current transaction with signer to the composer.
+    Then I simulate the current transaction group with the composer
+    And the simulation should report a failure at path "1" with message "transaction already in ledger"
+
+  @simulate
+  Scenario: Simulating bad inner transactions in the ATC
+    Given a new AtomicTransactionComposer
+    # Create an app at context index 0: FakeRandom
```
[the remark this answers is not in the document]
tzaffi: I agree that this is the biggest bang for the buck.

features/integration/simulate.feature
tzaffi: looks good

features/integration/simulate.feature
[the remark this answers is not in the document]
tzaffi: thx for issue #267 

resolving

features/integration/simulate.feature
```diff
@@ -0,0 +1,94 @@
...
+    And suggested transaction parameters from the algod v2 client
+    And I create a new transient account and fund it with 10000000 microalgos.
+    And I make a transaction signer for the transient account.
+
+  @simulate
+  Scenario Outline: Simulating successful payment transactions
+    Given default transaction with parameters <amt> "<note>"
+    When I get the private key
+    And I sign the transaction with the private key
+    And I simulate the transaction
```
tzaffi, line 15: Do you think it's worth distinguishing between simulating with signed transactions and simulating without ?

If so, this step could instead by:

```cucumber
And I simulate-signed the transaction
```
leaving room for the future

```cucumber
And I simulate-unsigned the transaction
```

But if the API is going to be the same (i.e. simulate just takes the transactions as they are and doesn't need special methods to handle unsigned), then we don't need to change this step.

We could also punt this to #267 and if the API changes at a later date, we would change the steps as well - though that would be a bit annoying.

features/integration/simulate.feature
```diff
@@ -0,0 +1,94 @@
...
+      | 0       | X4Bl4wQ9rCo= |
+      | 1234523 | X4Bl4wQ9rCo= |
+
+  @simulate
+  Scenario: Simulating duplicate transactions in a group and overspending errors
+    Given a new AtomicTransactionComposer
+    When I build a payment transaction with sender "transient", receiver "transient", amount 100001, close remainder to ""
+    And I create a transaction with signer with the current transaction.
+    And I add the current transaction with signer to the composer.
+    Then I simulate the current transaction group with the composer
```
tzaffi, line 29: Similar question. Should we anticipate the following step, and therefore rename this one?

```cucumber
Then I simulate-signed the current transaction group with the composer
```

tzaffi (commented):
Just two remaining questions: whether `source` is needed as `channel` in `.env` and whether our steps should assume that `simulate-signed` will have a different API than `simulate-unsigned`.

features/integration/simulate.feature
algochoi: I don't think the API would change, so need for two separate steps. 

Currently, the simulate endpoint itself sends the msgpack encoded transaction (bytearray-like object). As a part of #267 , we probably need helper methods to encode unsigned transactions into msgpack, and then send that to the simulate endpoint. Currently, we have explicit checks in the SDKs to look for signatures before doing this encoding.

tzaffi (approved):
LGTM

algochoi:
@tzaffi I added new tests for simulating unsigned transactions [a1023f8](https://github.com/algorand/algorand-sdk-testing/pull/265/commits/a1023f811540441cd571325588ed52a920321ed8) - let me know if you have any feedback!

algochoi:
Since we merged https://github.com/algorand/py-algorand-sdk/pull/420 and https://github.com/algorand/py-algorand-sdk/pull/445 in, it makes sense to merge this in as well.

PR #266, opened 2023-03-03
SANDBOX_CLEAN_CACHE with Default On

tzaffi:
Sometimes my local tests status mismatch what happens in C.I. One common culprit is Docker's layer caching. For example, recently, an indexer layer was being cached from December 2022. While tests were failing in C.I. they were passing on my machine.

This PR should fix this mismatch

This shouldn't impact C.I., as no layer caching should be taking place.

## What if I still want to cache sandbox layers?

You can turn layer caching back on by setting `SANDBOX_CLEAN_CACHE=0` in `.env`

### DIFF
--- a/.env
+++ b/.env
@@ -23,7 +23,7 @@ INDEXER_SHA=""
 # Sandbox configuration:
 SANDBOX_URL="https://github.com/algorand/sandbox"
 SANDBOX_BRANCH="master"
-SANDBOX_FORCE_RECREATE=1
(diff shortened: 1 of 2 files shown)


algochoi (approved):
This looks good to me based on a quick scan of `sandbox` here: https://github.com/algorand/sandbox/blob/master/sandbox#L237 

It looks like a `.clean` file will run docker compose with the `--no-cache` flag.

winder (approved):

PR #273, opened 2023-04-04
Enable all Indexer parameters

nullun:
We've been adding examples to the repos (e.g. [java-algorand-sdk](https://github.com/algorand/java-algorand-sdk/tree/examples/examples/src/main/java/com/algorand/examples)) that represent snippets we then include into the docs. Some of these use `currency-greater-than` but this fails the smoke test because the sandbox Indexer doesn't have it enabled.


bbroder-algo (approved):
fine w/ me up to will

bbroder-algo (approved):

PR #275, opened 2023-04-21
Add tests for devmode timestamps

algochoi:
Adds tests for get/setting timestamp offsets in dev mode. Only confirms that the algod request/response is correct as a sanity check.

Example here: https://github.com/algorand/py-algorand-sdk/pull/468

### DIFF
--- a/features/integration/devmode.feature
+++ b/features/integration/devmode.feature
@@ -1,0 +1,9 @@
+Feature: Dev mode features
+  Background:
+    Given an algod v2 client
+
+  @devmode.offset
+  Scenario: Setting timestamp offsets in dev mode
+    When I set the timestamp offset to be 1234
+    And I get the timestamp offset
+    Then the timestamp offset should be 1234
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ These reside in the [integration features directory](features/integration)
 | @c2c                   | Test Contract to Contract invocations and ingestion.                                   |
 | @compile               | Test the algod compile endpoint.                                                       |
 | @compile.sourcemap     | Test that the algod compile endpoint returns a valid Source Map.                       |
+| @devmode               | Test dev mode features, such as setting timestamp offsets.                             |
 | @dryrun                | Test the algod dryrun endpoint.                                                        |
 | @dryrun.testing        | Test the testing harness that relies on dryrun endpoint. Python only.                  |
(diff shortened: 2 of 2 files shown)


features/integration/devmode.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Dev mode features
+  Background:
+    Given an algod v2 client
+
+  @devmode.offset
+  Scenario: Setting timestamp offsets in dev mode
+    When I set the timestamp offset to be 1234
+    And I get the timestamp offset
+    Then the timestamp offset should be 1234
```
tzaffi, line 9: Is line 9 purely testing `GET /v2/devmode/blocks/offset/` ? If so, maybe it's also worth adding a step that calls `GET v2/block/{round}` and verifies that the offset had the desired effect? Might be tricky to make it non-flaky though...

tzaffi (commented):
Looks good. Just proposing an additional step (or two)

winder:
I don't think an integration test adds much safety here beyond whats already been done in the algod tests.

What do you think about adding the path/response unit tests and calling it a day?

For example: https://github.com/algorand/algorand-sdk-testing/pull/245/files

tzaffi:
> I don't think an integration test adds much safety here beyond whats already been done in the algod tests.
> 
> What do you think about adding the path/response unit tests and calling it a day?
> 
> For example: https://github.com/algorand/algorand-sdk-testing/pull/245/files

I think some kind of **E2E** in `go-algorand` or _integration_ test in one repo ought to exist showing that the devmode timestamps are behaving as expected. Maybe this already exists, but I didn't spot such a test in a brief search. Maybe add this test somewhere in [this file](https://github.com/algorand/go-algorand/blob/8fd6e702f2b319da5e404815f8af7a402e2a67f1/test/e2e-go/features/devmode/devmode_test.go#L33)?

algochoi:
Alternatively we could implement a full end to end test for devmode APIs for only one SDK (e.g. Python) under the @devmode integration tag and the rest can be tested with the path unit test?

algochoi:
> What do you think about adding the path/response unit tests and calling it a day?


Added path/response tests here: https://github.com/algorand/algorand-sdk-testing/pull/276

winder (changes requested):
Based on our previous conversations, I think this can be closed. Let me know if you disagree and we can discuss in person.

algochoi:
Sounds good, I'll close this for now.

PR #276, opened 2023-04-24
Add path tests for timestamp offset

algochoi:
Unit tests for timestamp offset endpoint and the ready endpoint.

Only GetBlockTimestampOffset returns a non-null response, so only that response test was updated.

Tested on the PySDK: https://github.com/algorand/py-algorand-sdk/pull/468


Eric-Warehime (approved):

winder:
Please add the new tag to the README. Could you also add the `/ready` endpoint?

winder (changes requested):
Please add the new tag to the README. Could you also add the `/ready` endpoint?

algochoi:
@winder 

Good point about README, I will fix that in the next commit - where should the `/ready` endpoint be added? It looks like version agnostic endpoint tests do not exist right now

winder:
@algochoi good point about the non-v2 test. How about `algodclient_paths.feature`?

features/unit/v2algodclient_paths.feature
```diff
@@ -179,3 +179,19 @@ Feature: Algod REST API v2 Paths
     Examples:
       | path            | method |
       | /v2/ledger/sync | delete |
+
+  @unit.timestamp
+  Scenario Outline: SetBlockTimeStampOffset path
+    When we make a SetBlockTimeStampOffset call against offset <offset>
+    Then expect the path used to be "<path>"
```
winder, line 186: May as well use the new one :shrug:
```suggestion
    Then expect the request to be "get" "<path>"
```

algochoi: Updated in : [9ec4a57](https://github.com/algorand/algorand-sdk-testing/pull/276/commits/9ec4a577a4df9c65d140dd8cc8c6ccbdd9f425c4)

PR #280, opened 2023-05-08
Low Resource Banning Error Message Change

ahangsu:


jasonpaulos (approved):
Looks correct, that error message did change in algorand/go-algorand#5328

PR #282, opened 2023-05-12
client: Add unit.client-no-headers tests

Eric-Warehime:
Adds tests which assert the headers being sent in requests when no auth token is given.

These are implemented for the JS SDK in https://github.com/algorand/js-algorand-sdk/pull/782

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ These reside in the [unit features directory](features/unit)
 | @unit.applications.boxes             | Box endpoints for Algod and Indexer.                                                                        |
 | @unit.atomic_transaction_composer    | ABI / atomic transaction construction unit tests.                                                           |
 | @unit.atc_method_args                | Test that algod's Atomic Transaction Composer assserts that the same number of arguments given as expected. |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -26,6 +26,7 @@ These reside in the [unit features directory](features/unit)
 | @unit.applications.boxes             | Box endpoints for Algod and Indexer.                                                                        |
 | @unit.atomic_transaction_composer    | ABI / atomic transaction construction unit tests.                                                           |
 | @unit.atc_method_args                | Test that algod's Atomic Transaction Composer assserts that the same number of arguments given as expected. |
+| @unit.client-no-headers              | Tests that clients don't add non-standard http headers when no (or an empty) auth token is provided.        |
```
tzaffi, line 29: Should this be Javascript only? If so, I recommend adding a row to the [not full supported table](https://github.com/algorand/algorand-sdk-testing/blob/master/README.md?plain=1#L83)

jasonpaulos (approved):

PR #286, opened 2023-05-19
algod: Update txn group delta for round response

Eric-Warehime:
Updates cucumber test responses based on output from https://github.com/algorand/go-algorand/pull/5404


algochoi (approved):
Looks good based on  [5404](https://github.com/algorand/go-algorand/pull/5404)

winder (approved):

PR #293, opened 2023-06-07
algod: Change groupdelta responses to msgp

Eric-Warehime:
Change responses to msgp format instead of json since the requests expect msgpack as the format supplied.


features/unit/responses.feature
```diff
@@ -217,11 +217,11 @@ Feature: REST Client Responses
     @unit.responses.txngroupdeltas
     Examples:
       | jsonfile                         | response-folder             | status | client | endpoint                                    |
-      | groupdelta-betanet_23963123.json | v2algodclient_responsejsons | 200    | algod  | GetTransactionGroupLedgerStateDeltaForRound |
```
tzaffi, line 220: I was expecting that `groupdelta-betanet_23963123.json` et al. would need to be removed from `features/unit/v2algodclient_responsejsons` but I didn't find them there. Was this a previous bug, or there's a different explanation?

Eric-Warehime: We can remove them if we want since they wouldn't be used anymore. I thought it might be nice to leave them in since they're relatively human readable equivalents of the msgp outputs. You can obviously use msgpacktool to be able to read them as well.

You think it's best to remove them?

tzaffi: I'm just not seeing them there so I'm confused:
```sh
❯ cd features/unit/v2algodclient_responsejsons

❯ ls
accountInformation.json          block.json                       empty.json                       pendingTransactions.base64       pendingTransactionsByTxid.base64 status.json
block.base64                     dryrunResponse.json              ledgerSupply.json                pendingTransactionsByAddr.base64 sendTx.json                      suggestedParams.json
```

Eric-Warehime: They're in `features/resources/v2algodclient_responsejsons/`
https://github.com/algorand/algorand-sdk-testing/blob/master/features/resources/v2algodclient_responsejsons/groupdelta-betanet_23963123.json

Not sure why we have different directories that are named the same...

tzaffi: Created an issue and approving.

Here's the issue #294

winder (approved):

PR #298, opened 2023-06-12
tests: Add json deltas, update sleep time

Eric-Warehime:
Two things in this PR

1. Adds json response tests for txn group deltas. Because of the issues w/ the json msgp decoder, we've decided to use the json response format in the JS SDK. This adds response tests for that under a new test tag.
2. Because of the recent sandbox changes we're now using conduit/indexer to ingest data from dev mode algod. It seems like this introduces slightly more delay in retrying block gets because there are some failures retrieving data in the application tests after the `500` millisecond sleeps. See [this go-algorand-sdk test run](https://app.circleci.com/pipelines/github/algorand/go-algorand-sdk/1020/workflows/e0d34709-a0b6-4050-89de-44e248ae3a19/jobs/2382) for example. I've double the wait time to ensure data propagates before the test runs.


bbroder-algo:
looks like good work thank you

bbroder-algo (approved):

PR #299, opened 2023-07-26
bug: Fix accountinfo response

Eric-Warehime:
The AccountInfo response that is provided does not provide some fields which are marked as required in the spec which causes errors when trying to decode this into an object in the JS SDK tests.


algochoi (approved):
Good catch, these look like mandatory fields in `Account`

shiqizng (approved):

Issue #182, opened 2022-05-26
`currencyGreaterThan` or `currencyLessThan` should appear in query if the value is explicitly set to 0


fionnachan:
### Subject of the issue
Various attempts were made to fix https://github.com/algorand/js-algorand-sdk/issues/414 but the PRs were blocked by failed cucumber tests which have an inaccurate expected behaviour for the feature.

See below
|PR author|PR|CircleCI test report|
|-----------|---|---------------------------|
| @algodoggo |https://github.com/algorand/js-algorand-sdk/pull/562|https://app.circleci.com/pipelines/github/algorand/js-algorand-sdk/158/workflows/9878838e-a782-48dc-89c6-0a74df888b9d/jobs/794|
| @barnjamin |https://github.com/algorand/js-algorand-sdk/pull/571|https://app.circleci.com/pipelines/github/algorand/js-algorand-sdk/180/workflows/971cbb96-78b7-4a19-91d2-1b17f4900151/jobs/925|

### Expected behaviour
These are the expected behaviour for the SDK.
1. [working as expected] When `currencyGreaterThan` or `currencyLessThan` is unset, it should not appear in the url query list.
2. [working as expected] When `currencyLessThan` is set to 0, the returned results should be empty or throw an error for disallowed value.
3. [working as expected] When `currencyGreaterThan` is set to 0, the returned result from the API is different from that of when `currencyGreaterThan` is unset. 
    **=> And that's why `currency-greater-than=0` should be in the query list if 0 is set explicitly.** 
4. **[not working]** When `currencyGreaterThan` or `currencyLessThan` is set to 0, it should appear in the url query list.

### Actual behaviour
There are many tests in https://github.com/algorand/algorand-sdk-testing/blob/master/features/unit/v2indexerclient_paths.feature that has the same inaccurate expectation. Below is one of the examples:
`currency-greater-than=0` or `currency-less-than=0` is expected to not appear in the url's query list by the cucumber tests even though `0` is being explicitly set.
![image](https://user-images.githubusercontent.com/13184582/170477678-b8e92467-1ec0-4196-a2e3-16ffa10350f3.png)
https://github.com/algorand/algorand-sdk-testing/blob/d60f7b5c6998ada2b755802a4c53b762063621e4/features/unit/v2indexerclient_paths.feature#L6-L15

AlgoDoggo:
[...]

PR #302, opened 2023-08-03
updating paths to include currencyGreaterThan param for 0

shiqizng:
#182. 
[py-algorand-sdk](https://github.com/algorand/py-algorand-sdk/pull/507) updates for these updated tests.


shiqizng:
@algochoi I had to reorder the params because json serialization in Go orders the fields alphabetically. Please have another look.

winder:
I think there are several other fields like this where the zero value is also a valid input

winder:
Maybe it's just GT and LT: https://github.com/algorand/indexer/pull/484

shiqizng:
> I think there are several other fields like this where the zero value is also a valid input

maybe, we need to do some investigation to know if there are other parameters having the same issue.

shiqizng:
> The main issue here is that it seems we can no longer distinguish between unset and set to 0. Is this an issue?

Unset and 0 are handled using None in python, String in java and js, pointer in Go, and 
we have some tests in here that have unset currencyGreaterThan values, https://github.com/algorand/algorand-sdk-testing/blob/48bee3e9cceb4d41a7a251f48914ed56ed73dcba/features/unit/v2indexerclient_paths.feature#L279. 

These tests remain unchanged, and all sdks are passing these tests.

algochoi (approved):
Looks good and targets the v2 branch correctly

winder (approved):

PR #307, opened 2024-01-19
Fix algod error string

jasonpaulos:
Fix algod error string, which changed because of https://github.com/algorand/go-algorand/pull/5875

### DIFF
--- a/features/integration/applications.feature
+++ b/features/integration/applications.feature
@@ -12,7 +12,7 @@ Feature: Applications
       Given I create a new transient account and fund it with 100000000 microalgos.
       # Application create with extra pages should succeed with expected message.
       And I build an application transaction with the transient account, the current application, suggested params, operation "create", approval-program "programs/big_app_program.teal.tok", clear-program "programs/four.teal.tok", global-bytes <global-bytes>, global-ints 0, local-bytes <local-bytes>, local-ints 0, app-args "", foreign-apps "", foreign-assets "", app-accounts "", extra-pages 3, boxes ""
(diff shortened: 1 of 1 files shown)


jannotti (approved):
This certainly seems fine, but I'm a little surprised it happened now.  Is there any chance this has been lurking for a fairly long time?  I'm a little unsure when these tests run.  This looks more closely related to https://github.com/algorand/go-algorand/pull/5336 than the more recent change.

jasonpaulos:
Fairly confident this was because of the structured error PR (5875), since this is the actual error string which caused a failure yesterday:

`TransactionPool.Remember: transaction 35OEHPL5FD7UJSAPFVFHQYEFSRNE6ROP3EDIZ6IZRFFAKF4JFIYQ: logic eval error: pc=704 dynamic cost budget exceeded, executing intc_1: local program cost was 700. Details: app=1072, pc=704, opcodes=intc_1 // 1; pop; intc_1 // 1`

from https://app.circleci.com/pipelines/github/algorand/js-algorand-sdk/1306/workflows/d935951b-4f64-4b80-a7ce-1a9a5de4ff51/jobs/7964

It's the `app=1072` which I believe is new and from that PR.

Though, to your point, these tests only run on SDK PRs/commits, so it possible for errors to be possible but not seen for a while. Before this error, circle says the last successful run on the JS SDK was 9 days ago though, so I don't think this error has been around for long.

jannotti:
I see, I misread the diff and was thinking you had to add the "Details" stuff, but you removed it. I guess this is a prefix match.

PR #308, opened 2024-02-22
Update indexer mock responses

jasonpaulos:
The existing responses are 4 years old and missing many new fields, and this PR aims to correct that. I produced new ones from a current indexer and I deleted files that were unused

- [x] Remove unused directory ` features/unit/v2algodclient_responsejsons`
- [x] Update responses in `features/resources/v2indexerclient_responsejsons`
- [ ] Update responses in `features/resources/generated_responses`


jasonpaulos:
It's no longer necessary to do this, since the JS SDK no longer panics if a required field is missing.

I've opened #315 to remove the unused files

